PR #10, opened 2018-12-05
Added google firebase authentication

IsaackMwenda:
Created a Login page
Added google firebase authentication
image


lukechurch:
I think you'll also need to adjust rules.json at the same time as this? You can do that by changing it in this PR

IsaackMwenda:
I have Added firestore.rules
Fixed page flickering on data update
Fixed identation and whitespaces
Added a firebase.js file containing firebase API statements
Added a favicon

IsaackMwenda:
Added columns for wrong scheme and not codable messages i.e (Number of messages and %)
image

lukechurch:
> Added columns for wrong scheme and not codable messages i.e (Number of messages and %)
> image

Thanks very much for this @IsaackMwenda, impressive! - however this is conceptually separate from authentication so in future on a separate branch and PR please. (cc @ritazagoni )

public/coda/firestore.rules
```diff
@@ -0,0 +1,16 @@
...
+        function getDatasetResource(){
+            return get(/databases/$(database)/documents/$(documents))
+        }
+        allow read: if userIsAllowed(resource)
+      match /metrics/{metrics=**} {
+      	allow read, write: if userIsAllowed(getDatasetResource())
+      }
+    }
+  }
+
```
lukechurch: Sorry @IsaackMwenda I don't think this works, I'm going to remove it from the PR

lukechurch (approved):

PR #11, opened 2018-12-09
Add rules files to ensure reads are authenticated

lukechurch:


IsaackMwenda:
@lukechurch  Thanks alot for this. I understand firebase authentication better now.

PR #19, opened 2019-05-03
Download and print the list of active projects from Firestore

as2388:


sms_statistics_service/sms_statistics_service.py
```diff
@@ -0,0 +1,38 @@
...
+from core_data_modules.logging import Logger
+
+from src import FirestoreClient
+
+Logger.set_project_name("OpsDashboard")
+log = Logger(__name__)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="TODO")
```
lukechurch, line 14: TODO

sms_statistics_service/src/firestore_client.py
```diff
@@ -0,0 +1,27 @@
+import firebase_admin
+from firebase_admin import credentials, firestore
+
+from src.data_models.active_project import ActiveProject
+
+
+class FirestoreClient(object):
```
lukechurch: I think this needs a different name - mainly because 'Client' is very heavily used by the Google API libraries.

Consider FirestoreWrapper

lukechurch (approved):

sms_statistics_service/sms_statistics_service.py
as2388: Implemented in #20

sms_statistics_service/src/firestore_client.py
as2388: changed

PR #20, opened 2019-05-03
Update the overall SMS totals for the active projects

as2388:
Review #19 first.

PR #22, opened 2019-05-06
Handle all message statuses

as2388:
Review #20 first, which requires #19

### DIFF
--- a/sms_statistics_service/sms_statistics_service.py
+++ b/sms_statistics_service/sms_statistics_service.py
@@ -77,13 +77,17 @@ if __name__ == "__main__":
         for msg in raw_messages:
             minute_stats = stats[msg.created_on.astimezone(pytz.utc).isoformat(timespec="minutes")]
 
+            # Message statuses are "documented" here:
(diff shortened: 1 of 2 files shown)


sms_statistics_service/sms_statistics_service.py
```diff
@@ -0,0 +1,104 @@
...
+        for msg in raw_messages:
+            minute_stats = stats[msg.created_on.astimezone(pytz.utc).isoformat(timespec="minutes")]
+
+            # Message statuses are "documented" here:
+            # https://github.com/rapidpro/rapidpro/blob/c972205aae29f7219582fc29478e8ecacb579f9f/temba/msgs/models.py#L79
+            if msg.direction == "in":
+                minute_stats.total_received += 1
+            elif msg.direction == "out":
+                if msg.status in {"initializing", "pending", "queued"}:
+                    pass  # Rapid Pro has not attempted to actually send the message yet, so don't increment any totals
```
lukechurch: I think I'd like these added to a counter. If we see a larger number of them it's a good warning that something is going wrong with the instance

lukechurch (approved):

sms_statistics_service/sms_statistics_service.py
as2388: changed to minute_stats.total_pending += 1

PR #27, opened 2019-05-13
Write counts to documents instead of document fields

as2388:


sms_statistics_service/src/firestore_wrapper.py
```diff
@@ -22,8 +22,8 @@ def __init__(self, cert):
     def _get_active_projects_collection_ref(self):
         return self.client.collection(f"active_projects")
 
-    def _get_sms_stats_doc_ref(self, project_name):
-        return self.client.document(f"metrics/rapid_pro/projects/{project_name}")
+    def _get_sms_stat_doc_ref(self, project_name, iso_string):
```
lukechurch: iso_string isn't very clear. Lots of things are 'iso' I'd prefer date_string or date_iso_string

sms_statistics_service/src/firestore_wrapper.py
```diff
@@ -47,14 +62,23 @@ def update_sms_stats(self, project_name, sms_stats_batch):
...
+                batch_counter = 0
 
-        if not doc_ref.get().exists:
-            log.warning(f"No SMS stats document exists for project {project_name}; will create one...")
-            doc_ref.set(updates)
-        else:
-            doc_ref.update(updates)
+        if batch_counter > 0:
+            batch.commit()
+            log.info(f"Final batch of {batch_counter} messages committed")
```
lukechurch, line 80: I usually reset the counter here just out of good house-keeping, and also that we can assert that the counter is zero at the end

sms_statistics_service/src/firestore_wrapper.py
```diff
@@ -47,14 +62,23 @@ def update_sms_stats(self, project_name, sms_stats_batch):
...
 
+        batch.commit()
```
lukechurch: This seems superfluous given ln 79-81?

sms_statistics_service/src/firestore_wrapper.py
```diff
@@ -47,14 +62,23 @@ def update_sms_stats(self, project_name, sms_stats_batch):
         :type sms_stats_batch: dict of str -> src.data_models.SMSStats
         """
         log.info(f"Batch updating {len(sms_stats_batch)} SMS stats for project {project_name}...")
+        batch = self.client.batch()
+        batch_counter = 0
+        total_progress = 0
```
lukechurch: I think total_counter is more consistent?

sms_statistics_service/src/firestore_wrapper.py
```diff
@@ -47,14 +62,23 @@ def update_sms_stats(self, project_name, sms_stats_batch):
...
+            batch.set(self._get_sms_stat_doc_ref(project_name, iso_string), sms_stats.to_dict())
+            batch_counter += 1
+            total_progress += 1
 
-        doc_ref = self._get_sms_stats_doc_ref(project_name)
-        updates = {iso_string: sms_stats.to_dict() for iso_string, sms_stats in sms_stats_batch.items()}
+            if batch_counter >= self.MAX_BATCH_SIZE:
+                batch.commit()
+                log.info(f"Batch of {batch_counter} messages committed, progress: {total_progress} / {len(sms_stats_batch)}")
+                batch = self.client.batch()
```
lukechurch: I don't know if you need to do this? I have a feeling you don't, could you check with one of e.g. the batch writing coda tools or the docs

lukechurch (approved):

sms_statistics_service/src/firestore_wrapper.py
as2388: Changed to `date_string`

sms_statistics_service/src/firestore_wrapper.py
as2388: removed

sms_statistics_service/src/firestore_wrapper.py
as2388: changed

sms_statistics_service/src/firestore_wrapper.py
as2388: I did this because that's what the firestore_client_wrapper does in the Coda 2 repo.

But I've just tested experimentally*, and you're right, I don't need this.

(* by running with >500 items, which succeeds)

sms_statistics_service/src/firestore_wrapper.py
as2388: done

PR #28, opened 2019-05-14
Write datetime as a Firestore field

as2388:
(In addition to the datetime in the id).

Review #27 first.


lukechurch:
Apart from comments before about iso_string as a var name, LGTM, thanks!

lukechurch (approved):

PR #32, opened 2019-05-17
Read from production db

IsaackMwenda:
Remove test firestore configurations
Read from production db, log incoming data & remove test plots code


public/media_engagement/media_engagement.js
```diff
@@ -34,335 +35,3 @@ firebase.auth().onAuthStateChanged(function (user) {
         window.location.replace('auth.html')
     }
 });
-
```
lukechurch, line 37: This PR appears to be deleting a lot of functionality as well as changing to production db. Can you confirm that is intended?

IsaackMwenda: Yes thats intention. The next pr covers the new dashboard design.

lukechurch: ok thanks, in general that should probably have gone into a separate PR as it's a different logical change

lukechurch (approved):

public/media_engagement/media_engagement.js
IsaackMwenda: Great got it.

PR #34, opened 2019-05-20
Engagement ui

IsaackMwenda:
Have separated the js and html files per project. Ideally the project links should be a dropdown in the navbar but currently not working due to jquery version. Am investigating this and will fix in a future pr. 
image


public/media_engagement/imaqal_media_engagement.js
```diff
@@ -0,0 +1,37 @@
+//Permorm Authentication then update data
```
lukechurch, line 1: s/Permorm/Perform

lukechurch (approved):

PR #42, opened 2019-06-04
Add a script for running the sms service incrementally

as2388:

PR #46, opened 2019-06-07
Revert "Add a script for running the sms service incrementally"

lukechurch:
Reverts AfricasVoices/Operations-Dashboard#42


lukechurch:
@as2388 

Actually this causes a bunch of problems.

Having to install dround on the virtual machines that we use to execute this doesn't seem like a good complexity tradeoff.

We could either set it up so that this was added as a dependency inside the docker container, or in the python script. Either would avoid having to do another custom setup step that's OS dependent on the context that we run these scripts from.

Sorry for merging the PR before realising this problem

as2388:
Pushing dround into the Docker container feels like the way to go to me (as we'd have to make some changes to the interface to the python script otherwise). I will do that.

PR #48, opened 2019-06-14
Run SMS Service Incrementally in Docker

as2388:
Attempt number 2 at this (previous was #42).

This pushes dround into a Docker container so no additional installs are needed on the host machine.


lukechurch (approved):
Thanks @as2388 !

PR #49, opened 2019-06-17
Fix path to credentials

lukechurch:
The credentials file is copied into the docker container prior to execution, however the script wasn't then using that file.

This change makes the script use that path in the same way as the non-incremental implementation


as2388 (approved):
(Sorry, I didn't find this in local testing because I only tested with an existing cache)

PR #50, opened 2019-07-02
Issue 38 log scale

ritazagoni:


lukechurch:
Hi @ritazagoni 

I don't quite follow the intention here. Is the submission of the _log version intended to be an experiment of 'the system with the ability to do log plotting' - with the intention that we would replace the existing file with it, or is the intention that we keep both files and switch between the chart forms by switching HTML file?

If the later - I'm a bit concerned about the code duplication, we already have way too much code duplication from the way in which these charts are built, rather than using templating, I'm a bit scared of adding any combinatorial expansion to that...

ritazagoni:
Sorry not being clear - the intention is to have one HTML file + js that do the switching, I just didn't want to break the existing setup, as I'm also thinking we should integrate the stacked bar view in switching and have a dropdown for log and stack (and linear if that still makes sense to keep, I'm not sure)

lukechurch:
Thanks @ritazagoni as discussed LGTM to merge and iterate

lukechurch (approved):

lukechurch:
@ritazagoni Shall we close this?

ritazagoni:
Agree, let's close it as y-axis capping is implemented.

lukechurch:
Thanks @ritazagoni

PR #51, opened 2019-07-02
Issue 37 stacked bar charts

ritazagoni:
Without legend for now, for 10 minute intervals. Kept failed as line as it's not disaggregated but will convert to standard bar chart.

### DIFF
--- a/public/media_engagement/imaqal_media_engagement_stacked.js
+++ b/public/media_engagement/imaqal_media_engagement_stacked.js
@@ -1,0 +1,271 @@
+//Perform Authentication then update data 
+firebase.auth().onAuthStateChanged(function (user) {
+    if (user) {
+        console.log("Attempting to bind: " + user.email)
+        const mediadb = firebase.firestore();
+        const settings = { timestampsInSnapshots: true };
+        mediadb.settings(settings);
+        var data = [];
+        mediadb.collection('/metrics/rapid_pro/IMAQAL/').onSnapshot(res => {
+            console.log(res)
+            // Update data every time it changes in firestore
+            res.docChanges().forEach(change => {
+
+                const doc = { ...change.doc.data(), id: change.doc.id };
+
+                switch (change.type) {
+                    case 'added':
+                        data.push(doc);
+                        break;
+                    case 'modified':
+                        const index = data.findIndex(item => item.id == doc.id);
+                        data[index] = doc;
+                        break;
+                    case 'removed':
+                        data = data.filter(item => item.id !== doc.id);
+                        break;
+                    default:
+                        break;
+                }
+            });
+            update(data);
+        });
+        console.log('Bind Successful');
+    } else {
+        window.location.replace('auth.html')
+    }
+});
+//Create margins for the two graphs
+const Margin = { top: 40, right: 100, bottom: 50, left: 70 };
+const Width = 900 - Margin.right - Margin.left;
+const Height = 500 - Margin.top - Margin.bottom;
+
+// Append total received sms graph to svg
+var total_received_sms_graph = d3.select(".total_received_sms_graph").append("svg")
+    .attr("width", Width + Margin.left + Margin.right)
+    .attr("height", Height + Margin.top + Margin.bottom)
+    .append("g")
+    .attr("transform",
+        "translate(" + Margin.left + "," + Margin.top + ")");
+
+// Append total sent sms graph to svg
+var total_sent_sms_graph = d3.select(".total_sent_sms_graph").append("svg")
+    .attr("width", Width + Margin.left + Margin.right)
+    .attr("height", Height + Margin.top + Margin.bottom)
+    .append("g")
+    .attr("transform",
+        "translate(" + Margin.left + "," + Margin.top + ")");
+
+// Append total sent sms graph to svg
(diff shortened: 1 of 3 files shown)


public/media_engagement/imaqal_media_engagement_stacked.js
```diff
@@ -0,0 +1,314 @@
...
+    data = data.filter(a => a.datetime > new Date("2019-07-01T12:30:00+00:00"));
+    
+    receivedKeys = ["NC_received", "golis_received", "hormud_received", "nationlink_received", "somnet_received", 
+    "somtel_received", "telegram_received", "telesom_received"]
+
+    let stackReceived = d3.stack()
+            .keys(receivedKeys)
+    let receivedDataStacked = stackReceived(data)
+
+    sentKeys = ["NC_sent", "golis_sent", "hormud_sent", "nationlink_sent", "somnet_sent",
```
lukechurch: I think we'll need to read these dynamically in future, but I see the existing code also has this problem, so feel free to file an issue rather than solving immediately.

public/media_engagement/imaqal_media_engagement_stacked.js
```diff
@@ -0,0 +1,314 @@
...
+    "somtel_sent", "telegram_sent", "telesom_sent"]
+
+    let stackSent = d3.stack()
+            .keys(sentKeys)
+    let sentdDataStacked = stackSent(data)
+
+
+    // set scale domains
+    x.domain(d3.extent(data, d => new Date(d.datetime)));
+    y_total_received_sms.domain([0, d3.max(data, function (d) { return Math.max(d.total_received, d.golis_received, d.hormud_received, d.nationlink_received, d.somnet_received, d.somtel_received, d.telesom_received); })]);
```
lukechurch, line 153: Why do we need to do this?

max of (total, ... list of things that sum to total ... )

Will just return total?

public/media_engagement/imaqal_media_engagement_stacked.js
```diff
@@ -0,0 +1,314 @@
...
+    y_total_sent_sms.domain([0, d3.max(data, function (d) { return Math.max(d.total_sent, d.golis_sent, d.hormud_sent, d.nationlink_sent, d.somnet_sent, d.somtel_sent, d.telesom_sent); })]);
+    y_total_failed_sms.domain([0, d3.max(data, function (d) { return Math.max(d.total_errored); })]);
```
lukechurch, line 155: this will just return the value it's computing the max of?

public/media_engagement/imaqal_media_engagement_stacked.js
```diff
@@ -0,0 +1,314 @@
...
+        d.hormud_sent= +d.operators["hormud"]["sent"]
+        d.nationlink_sent= +d.operators["nationlink"]["sent"]
+        d.somnet_sent= +d.operators["somnet"]["sent"]
+        d.somtel_sent= +d.operators["somtel"]["sent"]
+        d.telesom_sent= +d.operators["telesom"]["sent"]
+    });
+
+    data.sort((a, b) => new Date(a.datetime) - new Date(b.datetime));
+
+    data = data.filter(a => a.datetime > new Date("2019-07-01T12:30:00+00:00"));
```
lukechurch: Was this development code? I think we should probably remove it?

lukechurch (approved):
Thanks very much @ritazagoni ! Really great to see this progress.

I've left a couple of comments to address, we can then merge, thanks again!

lukechurch:
Once we've got this landed, please could we also replace the other projects data, and then rename these to not _stacked, so they replace the current chart HTML files

public/media_engagement/imaqal_media_engagement_stacked.js
ritazagoni: Thanks, issue created.

public/media_engagement/imaqal_media_engagement_stacked.js
ritazagoni: Makes sense, been reusing this from the original code and can't make it work any other way. `y_total_received_sms.domain([0, function (d) { return Math.max(d.total_received, d.golis_received, d.hormud_received, d.nationlink_received, d.somnet_received, d.somtel_received, d.telesom_received); }]);` returns `Nan` and can't use `d3.max()` with a list of values as accessors.

public/media_engagement/imaqal_media_engagement_stacked.js
ritazagoni: Need to limit view in for the 10min bars. Updated to render last two days.

ritazagoni:
Thanks @lukechurch! Comments partly addressed - are you OK with merging?

public/media_engagement/imaqal_media_engagement_stacked.js
ritazagoni: Added in latest version.

ritazagoni:
Added legend and dynamically populating operators. @lukechurch please give it another review then I'll update other projects and rename files for production. Sorry for piling changes in one PR.

lukechurch:
Thanks @ritazagoni  - I'm going to merge now.

PR #57, opened 2019-07-11
Control y axis

ritazagoni:


public/media_engagement/imaqal_media_engagement.js
```diff
@@ -362,5 +363,50 @@ const update = (data) => {
     .attr("text-anchor", "start")
     .style("fill", "blue")
     .text("Total Failed");
+
+    function updateReceivedChart() {
```
lukechurch: I think this function name might be a bit confusing - it's really responding to a limit, hence the reason why 'this' is meaningful. Consider updateRecievedChartLimit()

public/media_engagement/imaqal_media_engagement.js
```diff
@@ -362,5 +363,50 @@ const update = (data) => {
...
+        receivedLayer.selectAll('rect')
+            .data(function(d) { return d })
+            .attr('x', function (d) { return x(d.data.datetime) })
+            .attr('y', function (d) { return y_total_received_sms(d[1]) })
+            .attr('height', function (d) { return y_total_received_sms(d[0]) - y_total_received_sms(d[1]) })
+            .attr('width', Width / Object.keys(data).length)
+    
+    }
+
+    function updateSentChart() {
```
lukechurch: As above

lukechurch (approved):

Issue #52, opened 2019-07-02
Display last seen date on each chart


lukechurch:
At the moment if the system updating the chart goes down, there's no easy way of seeing it.

We should have a date string for 'last update seen' which is normally formatted small and black - and becomes large and red if the value is more than 20 minutes old.

IsaackMwenda:
@lukechurch should we close this?

as2388:
> @lukechurch should we close this?

This isn't working very well for me:
<img width="609" alt="image" src="https://user-images.githubusercontent.com/7963608/65978091-fa637880-e46a-11e9-8249-4df4eba3bcaf.png">
This is showing as being scary and red but it's been less than ten minutes since the last update so that doesn't make sense...

PR #63, opened 2019-08-15
Timestamp

ritazagoni:
Add "last updated field" as per issue #52.

### DIFF
--- a/public/media_engagement/imaqal_media_engagement.js
+++ b/public/media_engagement/imaqal_media_engagement.js
@@ -656,11 +656,11 @@ const update = (data) => {
         }
     });                        
     
-    var dayHourFormat = d3.timeFormat("%c")	
+    var fullDateFormat = d3.timeFormat("%c")	
 
     // Update timestamp of update and reset formatting
     const timestamp = new Date()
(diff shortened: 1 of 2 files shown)


public/media_engagement/imaqal_media_engagement.js
```diff
@@ -654,5 +654,25 @@ const update = (data) => {
...
+        var difference_ms = (currentTime.getTime() - timestamp.getTime())/60000
+        var difference_minutes = Math.floor(difference_ms % 60)
+     
+        // if updated more than 20 min ago >> reformat
+        if (difference_minutes > 20) {
+            d3.select("#lastUpdated").classed("alert", true)
+        }
+    }
+
+    setInterval(setLastUpdatedAlert, 1000)
```
lukechurch, line 677: Is this going to add new callbacks every time the data updates? If so we'll see a gradual accumulation of the callbacks over time?

lukechurch (approved):
LGTM with a question

public/media_engagement/imaqal_media_engagement.js
ritazagoni: Yes, good point. I tried to move the callback out of the data update function by making the function global defining it like window.setLastUpdateAlert, but then neither function was called. Any ideas?

PR #64, opened 2019-08-19
Update projects

ritazagoni:
Added timescale switch and y-axis limit control to OCHA and UNDP-RCO.


lukechurch:
LGTM - I would really like to get this templated so we don't have to do so much copying and pasting. Could we look at doing this with an HTML import in the first instance?

lukechurch (approved):

PR #65, opened 2019-08-27
Filter data

ritazagoni:
Filters data from firestore fetching only last 30 days. Also filters on client side. Currently implemented only for Imaqal, will add to other projects once they are updated.


lukechurch:
cc @DanielMwendwa

lukechurch (approved):

lukechurch:
cc @IsaackMwenda FYI

PR #78, opened 2019-10-10
Nav links

### LINKED ISSUE #71: Navbar: Use relative URLs
All the URLs in the nav-bar are currently hard-coded to firebaseapp.com. This makes testing in development environments a pain, and means the whole site will break if change the public URL on which it is hosted. We should switch to relative URLs.

DanielMwendwa:
Closes: #71


IsaackMwenda:
@DanielMwendwa as discussed offline, please test hosting the dashboard in a test firebase environment first then send re-review

PR #80, opened 2019-10-10
Ocha graphs

### LINKED ISSUE #77: Graphs Color scheme
Duplicate of issue #69. Most of the colors in the color palettes we are using are indistinguishable when they are used on the graphs, instead we should provide color names that we can differentiate on the graphs.

DanielMwendwa:
Closes: #77


DanielMwendwa:
Ocha graphs have been deleted

PR #86, opened 2019-10-25
removed all traffics except imaqal

DanielMwendwa:
I have also removed the traffics on the drop-down menu.


IsaackMwenda (approved):
Thanks @DanielMwendwa

lukechurch (approved):

PR #98, opened 2019-11-04
Worldbank collection

DanielMwendwa:
Changed db collection


as2388 (commented):
I'm not seeing messages from telegram on the graph, and there is a new 'deleted' item in the legend. Can you explain those or, even better if you've got round to addressing them, point me to the PR which fixes them?

DanielMwendwa:
> I'm not seeing messages from telegram on the graph, and there is a new 'deleted' item in the legend. Can you explain those or, even better if you've got round to addressing them, point me to the PR which fixes them?

@as2388, Messages from telegram were not displayed on the graph because the telegram operator was missing from the update data function. As for the deleted option I haven't found the reason behind it appearance. I also checked with Isaack it was not present in the collection. The legends shown on the graphs come from a set of operators therefore it means the deleted option is part of the operators but also I was unable to access it like other operators. Please check the collection for all the operators.

as2388 (approved):
@DanielMwendwa thanks for this and for fixing the telegram issue.

As discussed offline, the 'deleted' issue is due to a problem with the sms_statistics_service script, which needs to be fixed elsewhere.

lukechurch (approved):

PR #104, opened 2019-11-07
Worldbank x axis

### LINKED ISSUE #91: Width of xAxis for a bar chart too short
From our one day range drawn graphs our X domain only goes to XMax, which will be the bottom-left point of the last bar of the graph - it's not coming up "short", it just doesn't know how wide the bar is after XMax. Our bars' width is dependent on the number of keys of the daily received/sent total messages. The smaller the number of keys the wider the bars. 

(The XMax does not account for any additions made to the barWidth , and the text placement does account for the additional width, which seems to be why we're seeing the problem only with the Axis.)

In order to extend the axis all the way to the end, we'll have to account for the additional barWidth in our domain, or append the additional barWidth another way. That might be possible by adding another half day or so on to our xMax, or by shifting each bar to the left of its X position (-barWidth) instead of +barWidth.

DanielMwendwa:
Closes: #91

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
--- a/public/media_engagement/styles.css
+++ b/public/media_engagement/styles.css
@@ -24,7 +24,7 @@
 .txt-brown {
     color: #8d3538; 
 }
-.btn-brown{
+.btn-brown {
     border-color: #8d3538;
     color: #8d3538;
 }
--- a/public/media_engagement/styles.css
+++ b/public/media_engagement/styles.css
@@ -21,11 +21,14 @@
     background-color: #B6D4E7;
     
 }
-.btn-logout{
+.txt-brown {
+    color: #8d3538; 
+}
+.btn-custom{
     border-color: #8d3538;
     color: #8d3538;
 }
-.btn-logout:hover {
+.btn-custom:hover {
     color: steelblue;
     transform: scale(1.05);
 }
--- a/public/media_engagement/styles.css
+++ b/public/media_engagement/styles.css
@@ -24,11 +24,11 @@
 .txt-brown {
     color: #8d3538; 
 }
-.btn-custom{
+.btn-brown{
     border-color: #8d3538;
     color: #8d3538;
 }
-.btn-custom:hover {
+.btn-brown:hover {
     color: steelblue;
     transform: scale(1.05);
 }
--- a/public/media_engagement/worldbank_plr_sms_traffic.js
+++ b/public/media_engagement/worldbank_plr_sms_traffic.js
@@ -48,7 +48,12 @@ const TIMEFRAME_MONTH = 30;
 var chartTimeUnit = "10min";
 var isYLimitReceivedManuallySet = false;
 var isYLimitSentManuallySet = false;
-extend_x_max_by_hours = (date) => date.setHours(date.getHours() + 24);
+
+function add_one_day_to_date(day) {
+    var newDate = new Date(day);
+    newDate.setHours(newDate.getHours() + 24);
+    return newDate;
+}
 
 // Function to update data
 const update = (data) => {
@@ -192,7 +197,7 @@ const update = (data) => {
 
     // Set x and y scales
     const x = d3.scaleTime().range([0, Width]);
-    const x_axis_failed_sms_range = d3.scaleTime().range([0, Width]);
+    const failed_messages_x_axis_range = d3.scaleTime().range([0, Width]);
     const y_total_received_sms_range = d3.scaleLinear().range([Height, 0]);
     const y_total_sent_sms = d3.scaleLinear().range([Height, 0]);
     const y_total_failed_sms = d3.scaleLinear().range([Height, 0]);
@@ -228,7 +233,7 @@ const update = (data) => {
     // Define line paths for total failed sms(s)
     const total_failed_line = d3.line()
         .curve(d3.curveLinear)
-        .x(function (d) { return x_axis_failed_sms_range(new Date(d.datetime)) })
+        .x(function (d) { return failed_messages_x_axis_range(new Date(d.datetime)) })
         .y(function (d) { return y_total_failed_sms(d.total_errored); })
 
     // Create line path element for failed line graph
@@ -243,8 +248,8 @@ const update = (data) => {
     // set scale domain for failed graph
     y_total_failed_sms.domain([0, d3.max(data, function (d) { return d.total_errored; })]);
     xMin = d3.min(data, d => new Date(d.day));
-    xMax = d3.max(data, d => extend_x_max_by_hours(new Date(d.day)) )
-    x_axis_failed_sms_range.domain([xMin, xMax]);
+    xMax = d3.max(data, d => add_one_day_to_date(d.day)) 
+    failed_messages_x_axis_range.domain([xMin, xMax]);
 
     var yLimitReceived = d3.max(dailyReceivedTotal, function (d) { return d.total_received; });
     var yLimitReceivedFiltered = d3.max(dataFilteredWeek, function (d) { return d.total_received; });
@@ -287,7 +292,7 @@ const update = (data) => {
     //Add the X Axis for the total failed sms graph
     total_failed_sms_graph.append("g")
         .attr("transform", "translate(0," + Height + ")")
-        .call(d3.axisBottom(x_axis_failed_sms_range)
+        .call(d3.axisBottom(failed_messages_x_axis_range)
             .ticks(5)
             .tickFormat(timeFormat))
         // Rotate axis labels
@@ -505,7 +510,7 @@ const update = (data) => {
         }
 
         xMin = d3.min(data, d => new Date(d.day));
-        xMax = d3.max(data, d => extend_x_max_by_hours(new Date(d.day)) )
+        xMax = d3.max(data, d => add_one_day_to_date(d.day)) 
         // set scale domains
         x.domain([xMin, xMax]);
         y_total_received_sms_range.domain([0, yLimitReceived]);
@@ -657,7 +662,7 @@ const update = (data) => {
         }
 
         xMin = d3.min(data, d => new Date(d.day));
-        xMax = d3.max(data, d => extend_x_max_by_hours(new Date(d.day)) )
+        xMax = d3.max(data, d => add_one_day_to_date(d.day)) 
         // set scale domains
         x.domain([xMin, xMax]);
         y_total_sent_sms.domain([0, yLimitSent]);
--- a/public/media_engagement/worldbank_plr_sms_traffic.js
+++ b/public/media_engagement/worldbank_plr_sms_traffic.js
@@ -49,9 +49,9 @@ var chartTimeUnit = "10min";
 var isYLimitReceivedManuallySet = false;
 var isYLimitSentManuallySet = false;
 
-function add_one_day_to_date(day) {
-    var newDate = new Date(day);
-    newDate.setHours(newDate.getHours() + 24);
+function add_one_day_to_date(date) {
+    var newDate = new Date(date);
+    newDate.setDate(newDate.getDate() + 1);
     return newDate;
 }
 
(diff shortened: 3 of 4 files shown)


public/media_engagement/worldbank_plr_sms_traffic.js
```diff
@@ -45,10 +45,18 @@ firebase.auth().onAuthStateChanged(function (user) {
 
 const TIMEFRAME_WEEK = 7;
 const TIMEFRAME_MONTH = 30;
-var chartTimeUnit = "1day";
+const EXTEND_X_AXIS_BY_HOURS = 32
+var chartTimeUnit = "10min";
```
as2388, line 48: I like this, but I think it's an unrelated change that should have been in another PR

public/media_engagement/worldbank_plr_sms_traffic.js
```diff
@@ -45,10 +45,18 @@ firebase.auth().onAuthStateChanged(function (user) {
 
 const TIMEFRAME_WEEK = 7;
 const TIMEFRAME_MONTH = 30;
-var chartTimeUnit = "1day";
+const EXTEND_X_AXIS_BY_HOURS = 32
```
as2388: The use of the verb 'extend' makes this read like an action, or method name. As it's really a property, consider something like `X_AXIS_EXTENSION_HOURS`

public/media_engagement/worldbank_plr_sms_traffic.js
```diff
@@ -45,10 +45,18 @@ firebase.auth().onAuthStateChanged(function (user) {
...
+var chartTimeUnit = "10min";
 var isYLimitReceivedManuallySet = false;
 var isYLimitSentManuallySet = false;
 
+// Date objects inherit from Date.prototype
+// Add the addHours property to the date object
+Date.prototype.addHours= function(h){
+    this.setHours(this.getHours()+h);
+    return this;
```
as2388: Where possible, try to write functions that either return a value or mutate state only. Otherwise, if you mutate state and return a value it's really hard to predict which object(s) is being updated, which can lead to subtle bugs. 

So in this case, either call this.setHours and don't return any value, or create a new Date object by copying this one, add the hours to that, and return the copy

public/media_engagement/worldbank_plr_sms_traffic.js
```diff
@@ -45,10 +45,18 @@ firebase.auth().onAuthStateChanged(function (user) {
...
 const TIMEFRAME_MONTH = 30;
-var chartTimeUnit = "1day";
+const EXTEND_X_AXIS_BY_HOURS = 32
+var chartTimeUnit = "10min";
 var isYLimitReceivedManuallySet = false;
 var isYLimitSentManuallySet = false;
 
+// Date objects inherit from Date.prototype
+// Add the addHours property to the date object
+Date.prototype.addHours= function(h){
```
as2388: I recommend you don't modify the prototypes of library functions, particularly the standard library. Although you could probably get away with it here due to the small size of the project, in general it could cause problems - what if two libraries both attempted to add a different addHours for example? One of the implementations would be overwritten causing one of the libraries to break in a way which will be hard to debug.

Instead, create a function addHours(date: Date, hours: number): Date

public/media_engagement/worldbank_plr_sms_traffic.js
```diff
@@ -183,6 +191,7 @@ const update = (data) => {
 
     // Set x and y scales
     const x = d3.scaleTime().range([0, Width]);
+    const x_axis_failed_sms = d3.scaleTime().range([0, Width]);
```
as2388:
```suggestion
    const x_axis_failed_sms_range = d3.scaleTime().range([0, Width]);
```
or 
```suggestion
    const x_axis_failed_sms_scale = d3.scaleTime().range([0, Width]);
```
for added clarity

public/media_engagement/worldbank_plr_sms_traffic.js
```diff
@@ -183,6 +191,7 @@ const update = (data) => {
...
     const y_total_received_sms = d3.scaleLinear().range([Height, 0]);
```
as2388: Whatever you do with the x_axis, it would be nice to replicated here (e.g. y_axis_total_received_sms_scale).
(At the moment, I can't tell from the name what the type is. Appreciate this is a drive-by comment, possibly on someone else's code, but I think that would help)

public/media_engagement/worldbank_plr_sms_traffic.js
as2388: Also can you add a comment explaining the significance of the number 32?

24 would make sense because it's 1 day, but 32 seems very arbitrary to me.

as2388: I also wonder if a fixed constant is the most appropriate solution here. It has the side effect of extending the x-axis beyond the current day when in the 1 day view. Wouldn't it be better to extend the x-axis to the end of the current day?

as2388 (changes requested):
Thanks @DanielMwendwa. Requesting changes so I can take another quick look after you've addressed my comments

public/media_engagement/worldbank_plr_sms_traffic.js
DanielMwendwa: I used 32hours because it resolved the problem when I was working on it but I agree working with 24 because I also think that the data from operators also affect the width of the bars,  therefore the margin of the error was large before since I had not resolved the list of operators.

public/media_engagement/worldbank_plr_sms_traffic.js
DanielMwendwa: resolved

public/media_engagement/worldbank_plr_sms_traffic.js
DanielMwendwa: resolved

public/media_engagement/worldbank_plr_sms_traffic.js
DanielMwendwa: > I also wonder if a fixed constant is the most appropriate solution here. It has the side effect of extending the x-axis beyond the current day when in the 1 day view. Wouldn't it be better to extend the x-axis to the end of the current day?

By adding 24hours it extends the x-axis to the end of the current day since it accommodates the width of the bars

public/media_engagement/worldbank_plr_sms_traffic.js
DanielMwendwa: I didn't intend to make the change on this PR  but since it was approved on imaqal, I thought it was better to leave it there than to uncommit the changes.

as2388:
@DanielMwendwa I think you need to push this branch - seeing comments which are labelled 'resolved' but where the code is unchanged

public/media_engagement/styles.css
```diff
@@ -21,11 +21,14 @@
     background-color: #B6D4E7;
     
 }
-.btn-logout{
+.txt-brown {
+    color: #8d3538; 
+}
+.btn-brown{
```
as2388:
```suggestion
.btn-brown {
```

public/media_engagement/worldbank_plr_sms_traffic.js
```diff
@@ -45,9 +45,10 @@ firebase.auth().onAuthStateChanged(function (user) {
 
 const TIMEFRAME_WEEK = 7;
 const TIMEFRAME_MONTH = 30;
-var chartTimeUnit = "1day";
+var chartTimeUnit = "10min";
 var isYLimitReceivedManuallySet = false;
 var isYLimitSentManuallySet = false;
+extend_x_max_by_hours = (date) => date.setHours(date.getHours() + 24);
```
as2388: This modifies the date and returns it. Please do one or the other

as2388: Could you also change the function to name to something like 'add_one_day_to_date'?

 - By saying extend by 'hours' it sounds like users should be able to specify the number of hours, which they can't. 
 - This function's signature and implementation know nothing of x_max, so I think it's unnecessary to include that in the name. The function becomes much more general purpose with that name change.

public/media_engagement/worldbank_plr_sms_traffic.js
```diff
@@ -218,20 +228,23 @@ const update = (data) => {
...
-    color_scheme = ["#e6194B", "#f58231", "#3cb44b", "#CCCC00", "#4363d8", "#800000", "#f032e6", "#911eb4"]
+    color_scheme = ["#31cece", "#f58231", "#3cb44b", "#CCCC00", "#4363d8", "#800000", "#f032e6", "#911eb4", "#e6194B"]
     let color = d3.scaleOrdinal(color_scheme);
     let colorReceived = d3.scaleOrdinal(color_scheme).domain(receivedKeys);
     let colorSent = d3.scaleOrdinal(color_scheme).domain(sentKeys);
 
     // set scale domain for failed graph
     y_total_failed_sms.domain([0, d3.max(data, function (d) { return d.total_errored; })]);
+    xMin = d3.min(data, d => new Date(d.day));
+    xMax = d3.max(data, d => extend_x_max_by_hours(new Date(d.day)) )
```
as2388:
```suggestion
    xMax = d3.max(data, d => extend_x_max_by_hours(new Date(d.day)))
```

public/media_engagement/worldbank_plr_sms_traffic.js
```diff
@@ -274,7 +287,7 @@ const update = (data) => {
     //Add the X Axis for the total failed sms graph
     total_failed_sms_graph.append("g")
         .attr("transform", "translate(0," + Height + ")")
-        .call(d3.axisBottom(x)
+        .call(d3.axisBottom(x_axis_failed_sms_range)
```
as2388: Sorry to go round on this again, but on second reading today I think `failed_messages_x_axis_range` would read even better

as2388 (approved):
Thanks @DanielMwendwa, this makes the graphs look much cleaner/less confusing :)

Approved subject to comments being addressed, then over to Luke

public/media_engagement/worldbank_plr_sms_traffic.js
DanielMwendwa: I have written another function to avoid this, here: https://github.com/AfricasVoices/Operations-Dashboard/pull/104/commits/7096dda0f36484443029b955ae713b66b14b0c8d

public/media_engagement/worldbank_plr_sms_traffic.js
DanielMwendwa: https://github.com/AfricasVoices/Operations-Dashboard/pull/104/commits/7096dda0f36484443029b955ae713b66b14b0c8d

public/media_engagement/worldbank_plr_sms_traffic.js
DanielMwendwa: no worries :)

DanielMwendwa: I will rename attributes for y axis in another branch

public/media_engagement/worldbank_plr_sms_traffic.js
DanielMwendwa: @as2388 please take a look on this changes that you have requested

as2388: looks better.
 - Is day of type number or Date? The function name suggests it should be a date but the variable suggests a number - please update the function name or variable name appropriately
 - Does JS provide getDays and setDays functions that allow you to setDays(getDays() +1)? I think that would be more readable

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
as2388, line 1: Please don't commit this. Feel free to add .vscode/ to the .gitignore instead though

DanielMwendwa: Thank you, I will do so.

lukechurch:
@DanielMwendwa - please can you address the issues that @as2388 has raised before requesting a review from me? Thanks!

lukechurch (approved):
LGTM subject to @as2388 changes

public/media_engagement/worldbank_plr_sms_traffic.js
DanielMwendwa: Day is of type Date, so the function name is correct. I'll just change the variable name.

DanielMwendwa: Yes it does let me change to days. Thank you @as2388 this has helped a great deal.

DanielMwendwa:
> @DanielMwendwa - please can you address the issues that @as2388 has raised before requesting a review from me? Thanks!

@lukechurch the issue has been resolved.

PR #105, opened 2019-11-07
Label messages from deleted URNs as "NC"

as2388:


IsaackMwenda (approved):

lukechurch:
Can you message tech_data on slack with a brief note as to how we get to have deleted URNs?

lukechurch (approved):

sms_statistics_service/sms_statistics_service.py
```diff
@@ -99,6 +99,8 @@
             if msg.urn.startswith("tel:"):
                 # Set the operator name from the phone number
                 operator_name = PhoneCleaner.clean_operator(msg.urn)
+            elif msg.urn.startswith("deleted:"):
+                operator_name = Codes.NOT_CODED
```
lukechurch: NC or really actually missing?

as2388: As discussed offline, we'll change this to Codes.DELETED. 

This requires adding Codes.DELETED to Core, which is currently hard because there are outstanding PRs in Core that change how the control codes are used. To avoid implementing this twice because a PR is open, I propose delaying this PR until those are closed.

as2388:
I think we should hold this until we have https://github.com/AfricasVoices/Operations-Dashboard/issues/147, so that we don't break our existing colours by introducing this new code.

It's not urgent - we haven't deleted a contact this month so it's not actually visible on any graph.

PR #109, opened 2019-11-12
made changes on layout of the pages

DanielMwendwa:
Changed the layout of the pages in that the settings are displayed in bold, the card have a shadow and the message setting statement starts at the left side.


IsaackMwenda (approved):
Thanks @DanielMwendwa

PR #116, opened 2019-11-15
enabled user to logout of the dashboard

DanielMwendwa:


IsaackMwenda (approved):
Thanks @DanielMwendwa . Since this is a minimal change you can merge without re-questing re-review from @lukechurch.

PR #119, opened 2019-11-21
changed the sign in success url to imaqal page

DanielMwendwa:
The error when login into the dashboard after signing out from one of the sms traffic pages is encountered since the  'signInSuccessUrl' attribute was set to the earlier name of sms traffic page which was renamed for clarity, Therefore I have changed it to Imaqal page.

### DIFF
--- a/public/media_engagement/auth.html
+++ b/public/media_engagement/auth.html
@@ -11,7 +11,7 @@
     <script>
         // FirebaseUI config.
         var uiConfig = {
-            signInSuccessUrl: "imaqal_sms_traffic.html",
+            signInSuccessUrl: "../coda/coding_progress.html",
             signInOptions: [
             firebase.auth.GoogleAuthProvider.PROVIDER_ID
             ],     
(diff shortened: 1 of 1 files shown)


public/media_engagement/auth.html
```diff
@@ -11,7 +11,7 @@
     <script>
         // FirebaseUI config.
         var uiConfig = {
-            signInSuccessUrl: "media_engagement.html",
+            signInSuccessUrl: "imaqal_sms_traffic.html",
```
IsaackMwenda: I think it would be better to redirect back to `../coda/coding_progress.html` because the SMS traffic pages are not permanent.

IsaackMwenda (changes requested):

public/media_engagement/auth.html
IsaackMwenda: @DanielMwendwa  whats the progress on this?

DanielMwendwa: finished

IsaackMwenda (approved):
Thanks please merge

PR #122, opened 2019-11-30
modularized and refactored code

### LINKED ISSUE #75: Authentication and Data Binding
Since every media and engagement dashboard page is using the same function to authenticate and bind data. We should create one function that accepts user and traffic name as parameters and returns data of the given traffic, then import the function on every file that it's needed.

DanielMwendwa:
Closes #75 & #43 
Removed atleast 900 lines of  duplicate code


public/app.js
```diff
@@ -0,0 +1,330 @@
+// AUTH CONTROLLER
+var authController = (function() {
```
as2388: let?

public/app.js
```diff
@@ -0,0 +1,330 @@
...
+    return {
```
as2388: I don't write JS so much these days, but I believe you can just say:
```javascript
let authController = {
    initApp: function() {...},
    ...
};
```
i.e. no need to create and then call a function

public/app.js
```diff
@@ -0,0 +1,330 @@
...
+        //Authentication state listener
+        initApp: function() {
+            firebase.auth().onAuthStateChanged(function (user) {
+                if (user) {
+                    console.log('Login Successful');
```
as2388: (appreciate this may have been inherited from previous code, but if we could, standardise on either "" or '' please)

public/app.js
```diff
@@ -0,0 +1,330 @@
...
+                    console.log("Attempting to bind: " + user.email)
+                    console.log('Bind Successful');
```
as2388: I'm a bit confused as to what's going on here. You say you're attempted to bind a user than report that bind as successful without actually doing anything

public/app.js
```diff
@@ -0,0 +1,330 @@
+// AUTH CONTROLLER
+var authController = (function() {
+    return {
+        //Authentication state listener
```
```suggestion
        // Authentication state listener
```

public/app.js
```diff
@@ -0,0 +1,330 @@
...
+                    console.log("Attempting to bind: " + user.email)
+                    console.log('Bind Successful');
+                    return user;
+                } else {
+                    window.location.replace('auth.html')
+                }
+            });
+        },
+
+        //logout function
```
as2388: I don't think this comment is helpful. It's pretty obvious what `logout: function()` does and what its type is

public/app.js
```diff
@@ -0,0 +1,330 @@
+// AUTH CONTROLLER
+var authController = (function() {
+    return {
+        //Authentication state listener
+        initApp: function() {
```
as2388: Is this really initing the app? Or is it just getting a logged in user?

login() might actually be a better name, or get_user()

public/app.js
```diff
@@ -0,0 +1,330 @@
...
+            firebase.auth().onAuthStateChanged(function (user) {
```
as2388: We could use arrow functions here. Maybe worth a discussion in #team_data_tech about style here.

public/app.js
```diff
@@ -0,0 +1,330 @@
...
+                    window.location.replace('auth.html')
+                }
+            });
+        },
+
+        //logout function
+        logout: function() {
+            firebase.auth().signOut()
+                .catch(function (err) {
+                    console.log(err);
```
as2388: I think we should be much more noisy than just a console log. Maybe an alert() for now?

public/app.js
```diff
@@ -0,0 +1,330 @@
...
+                })
+        }
+    };
+})();
+
+// DATA CONTROLLER
+var dataController = (function() {
+    var data = [];
```
as2388: (note for other reviewers: this block to the end of the updateData function is copied unmodified from master)

as2388 (changes requested):
Thanks @DanielMwendwa. I really like where this is going but this is a massive change which re-architects the entire website. I reviewed the new authController so you can make progress, but would it be possible to break the rest of this PR into smaller, more manageable chunks?

public/app.js
DanielMwendwa: My idea was to create the functions and then call them on the global app controller where the logic of the dashboard is. What do you think about that?

as2388: I think I see what you're doing now. Consider:
```javascript
class AuthController {
    initApp() {
        ...
    }

    ...
}
```
Then yes, in the global controller you can say `AuthController.initApp()`.

DanielMwendwa: The AuthController will be undefined because it will not have been executed

DanielMwendwa:
> Thanks @DanielMwendwa. I really like where this is going but this is a massive change which re-architects the entire website. I reviewed the new authController so you can make progress, but would it be possible to break the rest of this PR into smaller, more manageable chunks?

I have finished the implementation of the entire application considering your idea of avoiding the use of IIFEs means restructuring the whole app again

public/app.js
DanielMwendwa: okay will go with double quotes

public/app.js
DanielMwendwa: I'll remove the comment

as2388:
> > Thanks @DanielMwendwa. I really like where this is going but this is a massive change which re-architects the entire website. I reviewed the new authController so you can make progress, but would it be possible to break the rest of this PR into smaller, more manageable chunks?
> 
> I have finished the implementation of the entire application considering your idea of avoiding the use of IIFEs means restructuring the whole app again

Sorry, I didn't realise the impact would be so extreme. But if you're about to restructure the whole app again then now would be a good time to do it in small pieces :)

public/app.js
as2388: works for me: https://jsbin.com/davalofebu/edit?js,console

DanielMwendwa: Noted your implementation, thanks

public/app.js
DanielMwendwa: As discussed earlier this is going to be a class

public/app.js
DanielMwendwa: Okay

public/app.js
DanielMwendwa: resolved

public/app.js
DanielMwendwa: Its getting a logged in user. I'll go with get_user for clarity

public/app.js
DanielMwendwa: noted

DanielMwendwa:
Thank you @as2388 for this review, I have implemented the changes we had discussed. I'll open a new PR where I'll present the changes in small pieces for review.

DanielMwendwa:
> Thank you @as2388 for this review, I have implemented the changes we had discussed. I'll open a new PR where I'll present the changes in small pieces for review.

Find the PR here #124

DanielMwendwa:
Issue resolved

PR #124, opened 2020-01-06
modularized and refactored code2

DanielMwendwa:


public/auth.html
```diff
@@ -0,0 +1,32 @@
...
+    <link rel="stylesheet" type="text/css" href="styles.css">
+    <script src="https://www.gstatic.com/firebasejs/5.5.9/firebase-app.js"></script>
+    <script src="https://www.gstatic.com/firebasejs/5.6.0/firebase-auth.js"></script>
+    <script src='firebase.js'></script>
+    <script src="https://cdn.firebase.com/libs/firebaseui/3.4.1/firebaseui.js"></script>
+    <link type="text/css" rel="stylesheet" href="https://cdn.firebase.com/libs/firebaseui/3.4.1/firebaseui.css" />
+    <script>
+        // FirebaseUI config.
+        var uiConfig = {
+            signInSuccessUrl: "index.html",
```
DanielMwendwa, line 14: I have changed this to point to index page instead of having two auth pages one pointing to coding progress page and the other to one of the media engagement page.

DanielMwendwa:
on e3610ef I have merged the two css files and removed duplicate code

public/data.js
```diff
@@ -0,0 +1,53 @@
...
+                case "modified":
+                    const index = input.findIndex(item => item.id == doc.id);
+                    input[index] = doc;
+                    break;
+                case "removed":
+                    input = input.filter(item => item.id !== doc.id);
+                    break;
+                default:
+                    break;
+            }
```
DanielMwendwa: this block is copied unmodified from master

DanielMwendwa:
On 8871f40 I have modified one of the media engagement js file to contain the code required to plot the graphs then I have put it in a class. The only change made on the code is while accessing the static method "add_one_day_to_date" and use of the let keyword to provide block scope variables otherwise the remaining code remains unmodified from master.

public/UI.js
```diff
@@ -0,0 +1,148 @@
...
+                            <th scope="col">Wrong Scheme messages</th>
+                            <th scope="col">WS %</th>
+                            <th scope="col">Not Coded messages</th>
+                            <th scope="col">NC %</th>
+                        </tr>
+                    </thead>
+                    <tbody id="coding_status_body"></tbody>
+                </table>
+            <div id="last_update">Last updated: </div>
+        </div> `
```
DanielMwendwa: this block is copied unmodified from master

public/UI.js
```diff
@@ -0,0 +1,148 @@
...
+                    }
+                    else if ((parseFloat(Done) > 50) && (parseFloat(Done) <= 75)) {
+                        $(this).parent().addClass('coding-above50');
+                    }
+                    else if ((parseFloat(Done) > 75) && (parseFloat(Done) < 100)) {
+                        $(this).parent().addClass('coding-above75');
+                    }
+                    else {
+                        $(this).parent().addClass('coding-complete');
+                    }
```
DanielMwendwa: this block is copied unmodified from master

public/UI.js
```diff
@@ -0,0 +1,148 @@
...
+            </section> 
+            <section>
+                <div class="d-md-flex justify-content-start mt-4 mb-3">
+                    <span class="font-weight-bold" type="text">Set the maximum number of outgoing messages you want to see</span> 
+                    <div class="col-md-2"><input class="form-control form-control-sm" type="number" id="buttonYLimitSent" step="500" min="10"></div>
+                </div>
+                <div class="card shadow total_sent_sms_graph"></div>
+            </section> 
+            <div class="card shadow total_failed_sms_graph my-4"></div> 
+        </div> `
```
DanielMwendwa: this block is copied unmodified from master

DanielMwendwa:
On 0e98548, the page replaces the media engagement HTML pages used for every project. The index page is dynamic as it is able to navigate to all the projects via the drop down menu. Some HTML code has been excluded here which will be handled here 567efef

DanielMwendwa:
I haven't deleted coda & media engagement folders to prevent merge conflicts in case of commits on the master branch & also for reference.

public/styles.css
```diff
@@ -0,0 +1,202 @@
...
+    font-size: 20px;
+}
+.container {
+    margin: 0 auto;
+    max-width: 1280px;
+    width: 85%;
+}
+@media only screen and (min-width: 601px) {
+    .container {
+        width: 80%;
```
as2388: Do we need this rule? Previously I could see the whole table on my laptop screen, now I have to horizontal scroll which isn't so great.

public/styles.css
```diff
@@ -0,0 +1,202 @@
...
+    }
+}
+@media only screen and (min-width: 993px) {
```
as2388: This rule is unnecessary due to the previous one

public/styles.css
```diff
@@ -0,0 +1,202 @@
+.logo {
+    height: 3rem;
+}
+.navbar {
+    box-shadow: 0 2px 2px -2px steelblue;
+}
+.navbar-brown{
```
as2388:
```suggestion
.navbar-brown {
```

public/styles.css
```diff
@@ -0,0 +1,202 @@
...
+}
+.active-link {
+    border-bottom: 2px #8d3538 solid;
+}
+.active-item {
+    color: #8d3538;
+    background-color: #B6D4E7;
+    
+}
+.btn-logout{
```
as2388:
```suggestion
.btn-logout {
```
etc.

public/styles.css
```diff
@@ -0,0 +1,202 @@
...
+.nav-link:hover {
+    color: steelblue;
+}
+.active-link {
+    border-bottom: 2px #8d3538 solid;
+}
+.active-item {
+    color: #8d3538;
+    background-color: #B6D4E7;
+
```
as2388:
```suggestion
```

public/UI.js
```diff
@@ -0,0 +1,148 @@
+// UI CONTROLLER
+class UIController {
+    static getDOMstrings() {
+        let DOMstrings = {
+            projectMenu: '.select__project',
```
as2388: Is there any reason these need to be different? If projectMenu is more descriptive, then why not change the css selector to `.project-menu`?

public/UI.js
```diff
@@ -0,0 +1,148 @@
+// UI CONTROLLER
+class UIController {
+    static getDOMstrings() {
+        let DOMstrings = {
```
as2388: I don't think separating the variable assignment from the return helps here, because the variable matches the function name.

Consider `return {...`

public/UI.js
```diff
@@ -0,0 +1,148 @@
...
+            codingProgressLink: '.codingprogress__link',
+            codingProgressContainer: '.codingprogress__table',
+            graphContainer: '.graph__list',
+            logoutBtn: '.logout__btn',
+        };
+        return DOMstrings;
+    }
+
+    static addDropdownMenu(data) {
+        let html, newHtml, DOM = UIController.getDOMstrings();
```
as2388: This pattern declares html and newHtml earlier than they are needed, and sets them to null or undefined (can't remember which), which is fine in this small case but can in general lead to harder-to-debug bugs, less readable code, and more complex types. I suggest only declaring variables when they get defined i.e. change to:
```
let DOM = UIController.getDOMstrings();
let html = ...
data.forEach(obj => {
    let newHtml = ...
```

as2388: getDOMstrings returns a DOMstrings object not the DOM itself, so better to call this variable DOMstrings.

public/UI.js
as2388: thanks for all these useful comments, they speed up the review :)

public/app.js
```diff
@@ -0,0 +1,57 @@
...
+        document.querySelector(DOM.codingProgressLink).addEventListener('click', Controller.ctrlDisplayCodingProgress);
+        document.querySelector(DOM.projectMenu).addEventListener('click', Controller.ctrlDisplayProject);          
+    };
+    
+    static ctrlLogoutDashboard() {
+        AuthController.logout()
+    };
+
+    // Navigate to coding progress page
+    static ctrlDisplayCodingProgress(e) {
```
```suggestion
    static navigateToCodingProgress(e) {
```
Then you don't need a comment explaining your function name :)

public/app.js
```diff
@@ -0,0 +1,57 @@
+// GLOBAL APP CONTROLLER
+class Controller {
+
```
```suggestion
```

public/auth.js
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    static logout() {
+        firebase.auth().signOut()
+            .catch(err => {
+                alert("An error occurred")
+                console.log(err);
+            })
+    }
+}
```
as2388:
```suggestion
}

```

public/auth.js
```diff
@@ -0,0 +1,20 @@
+class AuthController {
+    // Authentication state listener
+    static get_user() {
```
as2388:
```suggestion
    static getUser() {
```
(to follow JS naming convention)

public/UI.js
```diff
@@ -0,0 +1,148 @@
+// UI CONTROLLER
+class UIController {
+    static getDOMstrings() {
+        let DOMstrings = {
+            projectMenu: '.select__project',
+            codingProgressLink: '.codingprogress__link',
```
as2388: Do these need to be wrapped in an object in a function, or would it be cleaner to have something like:
```
class UIController {
    projectMenuSelector: "project-menu",
    codingProgressLinkSelector: ".coding-progress-link",
    ...,
    static addDropdownMenu(...
```

public/index.html
```diff
@@ -0,0 +1,58 @@
...
+    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.0.13/css/all.css" integrity="sha384-DNOHZ68U8hZfKXOrtjWvjxusGo9WQnrNx2sqG0tfsghAvtVlRW3tvkXWZh58N9jp" crossorigin="anonymous">
+    <link rel="stylesheet" href="styles.css">
+    <script src="firebase.js"></script>
+</head>
+<body>
+    <!-- NAVBAR WITH DROPDOWN & RESPONSIVE TOGGLE-->
+    <nav class="navbar navbar-expand-sm navbar-brown bg-white mb-1 py-0">
+        <div class="container">
+            <a class="navbar-brand" href="#"><img class="logo" src="avfdashboardlogo.png" alt="avf-logo"></a>
+            <!-- TOOGLE TRIGGER FOR SMALL DEVICES -->
```
as2388:
```suggestion
            <!-- TOGGLE TRIGGER FOR SMALL DEVICES -->
```

public/graph.js
```diff
@@ -0,0 +1,745 @@
+// GRAPH CONTROLLER
```
as2388, line 1: Can you let me know which of this is new?

public/index.html
```diff
@@ -0,0 +1,58 @@
...
+    <section class="graph__list"></section>
+    <script src="https://d3js.org/d3.v5.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/d3-legend/2.25.6/d3-legend.min.js"></script>
+    <script src="auth.js"></script>
+    <script src="data.js"></script>
+    <script src="UI.js"></script>
+    <script src="graph.js"></script>>
+    <script src="app.js"></script>
+</body>
+</html>
```
as2388:
```suggestion
</html>

```

public/data.js
```diff
@@ -0,0 +1,53 @@
+class DataController {
+    static updateData(response, input) {
```
as2388: Can you give the function parameters more descriptive names?

public/data.js
```diff
@@ -0,0 +1,53 @@
...
+                case "removed":
+                    input = input.filter(item => item.id !== doc.id);
+                    break;
+                default:
+                    break;
+            }
+        });
+    }
+
+    static getProject(update) {
```
as2388: It wasn't immediately obvious that `update` was should be a callback. I don't think calling this getProject helped either. Consider `watchActiveProjects(onChange)`

public/data.js
```diff
@@ -0,0 +1,53 @@
...
+        let activeProjects = []
+        mediadb.collection('active_projects').onSnapshot(res => {
+            // Update data every time it changes in firestore
+            DataController.updateData(res, activeProjects);
+            update(activeProjects);
+        });  
+    }
+
+    static getDocument(update) {
```
as2388: which document?

public/data.js
```diff
@@ -0,0 +1,53 @@
...
+        mediadb.doc('metrics/coda').onSnapshot(res => {
+            update(res.data());
+        });
+    }
+
+    static getCollection(collection, update) {
```
as2388: which collection?

public/app.js
```diff
@@ -0,0 +1,57 @@
...
+    };
+    
+    static ctrlLogoutDashboard() {
+        AuthController.logout()
+    };
+
+    // Navigate to coding progress page
+    static ctrlDisplayCodingProgress(e) {
+        if(e.target && e.target.nodeName == "A") {
+            window.location.reload();
```
as2388: I'm ok with this for now but add an issue for implementing a less aggressive navigation strategy

as2388 (changes requested):
Thanks @DanielMwendwa, this is really good progress. Requesting changes because the PR is quite large so I should check those changes when you're done.

A couple of other points I didn't know where else to add:
 - Can you make the project drop-down show the 'hand' icon when mousing over each project name in that list?
 - Can you make the project drop-down fit the height of the items in the list. At the moment there is some extra space which, when clicked, sets the timestamps to NaN and the project name to undefined.

public/UI.js
DanielMwendwa: This brings an error suggesting that the syntax applies to typescript files also public field declarations are an experimental feature and their support in browsers is limited.

public/auth.html
DanielMwendwa: thanks

public/styles.css
DanielMwendwa: resolved

public/styles.css
DanielMwendwa: resolved

public/data.js
DanielMwendwa: resolved

public/app.js
DanielMwendwa: done

public/data.js
DanielMwendwa: updated

public/UI.js
DanielMwendwa: updated

DanielMwendwa:
Thanks @as2388 for the comments, I have been able to make the requested changes.

public/app.js
```diff
@@ -0,0 +1,61 @@
...
+        document.querySelector(DOMstrings.projectMenu).addEventListener('click', Controller.navigateToSelectedProject);          
+    };
+
+    static resetDashboard() {
+        let DOMstrings = UIController.getDOMstrings();
+        document.querySelector(DOMstrings.codingProgressContainer).innerHTML = "";
+        document.querySelector(DOMstrings.graphContainer).innerHTML = "";
+    }
+    
+    static ctrlLogoutDashboard() {
```
as2388: This feels unnecessary. How about setting this directly in the event listener set-up i.e.
`document.querySelector(DOMstrings.logoutBtn).addEventListener('click', AuthController.logout);`

public/data.js
```diff
@@ -0,0 +1,54 @@
+class DataController {
+    static updateData(snapshot, data) {
+        // Update data every time it changes in firestore
+        snapshot.docChanges().forEach(change => {
+
```
as2388:
```suggestion
```

public/styles.css
```diff
@@ -0,0 +1,194 @@
...
+.table-heading {
+    background-color: #8e3436;
+    height: 30px;
+    color:#ffff;
+}
+.table thead > tr > th {
+    vertical-align: bottom;
+    border-bottom: 2px solid #8e3436;
+}
+table > tbody > tr > td{
```
as2388:
```suggestion
table > tbody > tr > td {
```

public/styles.css
```diff
@@ -0,0 +1,194 @@
...
+    border-bottom: 1px solid rgb(0,0,0, 0.2);
+}
+.coding-notstarted{
```
as2388:
```suggestion
.coding-notstarted {
```
and all other cases like this, I'm not going to manually highlight them all. Your code editor can probably be configured to help out here

public/index.html
```diff
@@ -0,0 +1,58 @@
...
+     <!-- Display Coding Progress Table Here -->
+    <section class="coding_progress coding-progress-container"></section>
+    <!-- Display Graphs Here -->
+    <section class="graph-container"></section>
+    <script src="https://d3js.org/d3.v5.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/d3-legend/2.25.6/d3-legend.min.js"></script>
+    <script src="auth.js"></script>
+    <script src="data.js"></script>
+    <script src="UI.js"></script>
+    <script src="graph.js"></script>>
```
as2388:
```suggestion
    <script src="graph.js"></script>
```

public/data.js
```diff
@@ -0,0 +1,54 @@
...
+        });  
+    }
+
+    static watchCodingProgress(onChange) {
+        mediadb.doc('metrics/coda').onSnapshot(res => {
+            onChange(res.data());
+        });
+    }
+
+    static watchProjectData(projectName, onChange) {
```
as2388: project coding data or project traffic data?

Suggest renaming this to 'watchProjectTrafficData' or similar

public/app.js
```diff
@@ -0,0 +1,61 @@
+// GLOBAL APP CONTROLLER
+class Controller {
+    static setupEventListeners() {
+        let DOMstrings = UIController.getDOMstrings();
+        document.querySelector(DOMstrings.logoutBtn).addEventListener('click', Controller.ctrlLogoutDashboard);
+        document.querySelector(DOMstrings.codingProgressLinkSelector).addEventListener('click', Controller.navigateToCodingProgress);
+        document.querySelector(DOMstrings.projectMenu).addEventListener('click', Controller.navigateToSelectedProject);          
+    };
+
+    static resetDashboard() {
```
```suggestion
    static resetUI() {
```
?

as2388: or clearUI?

public/app.js
```diff
@@ -0,0 +1,61 @@
...
+        document.querySelector(DOMstrings.codingProgressContainer).innerHTML = "";
+        document.querySelector(DOMstrings.graphContainer).innerHTML = "";
+    }
+    
+    static ctrlLogoutDashboard() {
+        AuthController.logout()
+    };
+
+    static navigateToCodingProgress(e) {
+        Controller.resetDashboard()
```
as2388: Should this be inside the if() statement?

public/app.js
```diff
@@ -0,0 +1,61 @@
...
+            // Add the coding progress section to the UI
+            UIController.addCodingProgressSection();
+            // Get data for coding progress table
+            DataController.watchCodingProgress(UIController.update_progress_ui);
+        }
+    };
+    
+    static navigateToSelectedProject(e) {
+        if(e.target && e.target.nodeName == "A") {
+            Controller.resetDashboard()
```
as2388: or this outside?

public/app.js
```diff
@@ -0,0 +1,61 @@
...
+        DataController.watchActiveProjects(UIController.addDropdownMenu);
+        // Add the coding progress section to the UI
+        UIController.addCodingProgressSection();
+        // Get data for coding progress table
+        DataController.watchCodingProgress(UIController.update_progress_ui);
+    } 
+} 
+// initialize firestore
+const mediadb = firebase.firestore();
+const settings = { timestampsInSnapshots: true };
```
as2388: In researching what this does I've discovered it is deprecated and can be removed:
https://firebase.google.com/docs/reference/js/firebase.firestore.Settings.html#timestampsinsnapshots

public/app.js
```diff
@@ -0,0 +1,61 @@
...
+             // Add the graphs container to the UI
+            UIController.addGraphs(project);
+            // Update and show the Graphs
+            DataController.watchProjectData(project, GraphController.update_graphs);
+        }
+    };  
+
+    static init() {
+        console.log('Application has started.');
+        // authorize user
```
```suggestion
        // Authorize user
```
(for consistency with your other comments)

as2388 (changes requested):
Thanks again for the improvements.

Two additional things:
 - I don't think you're turning off the snapshot listeners when navigating, which means as you navigate between views you end up activating lots of listeners you no longer need, which could cause bugs or performance issues. Can you check this?
 - By making this a one-page webapp, we've lost the ability to deep-link to particular pages, e.g. to link directly to the traffic page for a particular project. Could you add fixing that as an issue?

public/styles.css
DanielMwendwa: okay, thanks

public/app.js
DanielMwendwa: updated

public/app.js
DanielMwendwa: outside.

public/data.js
DanielMwendwa: Done

public/app.js
DanielMwendwa: I'm getting this warning after removing the setting which reverts me to adding the setting. 
firestore
We are also using Timestamp values instead of Date as suggested in the documentation. 
It is also stated that "To opt into the old behavior of returning Date objects, you can temporarily set timestampsInSnapshots to false." in our case it is true.

as2388: Ah, I suspect that's because we're using an old version of the client library. Maybe turn this into an issue instead

DanielMwendwa: Okay. Opened issue #128

DanielMwendwa: Opened issue #128

public/styles.css
```diff
@@ -0,0 +1,194 @@
...
+    background:#8d3538;
+    color: #fff;
+    top: 30%;
+    left: 50%;
+    position: absolute;
+    transform: translate(-50%,-50%);
+    box-sizing: border-box;
+    padding: 70px 30px;
+}
+.avatar{
```
as2388:
```suggestion
.avatar {
```

public/styles.css
```diff
@@ -0,0 +1,194 @@
...
+}
+.login-body{
+    margin: 0;
+    padding: 0;
+    background:#ffff;
+    background-size: cover;
+    background-position: center;
+    font-family: sans-serif;
+}
+.login-box{
```
as2388:
```suggestion
.login-box {
```

as2388 (changes requested):
Can't see the message coding table in Chrome 79:

image

DanielMwendwa:
> Thanks again for the improvements.
> 
> Two additional things:
> 
>     * I don't think you're turning off the snapshot listeners when navigating, which means as you navigate between views you end up activating lots of listeners you no longer need, which could cause bugs or performance issues. Can you check this?
Updated code based on this: https://firebase.google.com/docs/firestore/query-data/listen#detach_a_listener
> 
>     * By making this a one-page webapp, we've lost the ability to deep-link to particular pages, e.g. to link directly to the traffic page for a particular project. Could you add fixing that as an issue? 
yes, will open an issue for that.

DanielMwendwa:
> Can't see the message coding table in Chrome 79:
> 
> image

resolved

DanielMwendwa:
> > Thanks again for the improvements.
> > Two additional things:
> > ```
> > * I don't think you're turning off the snapshot listeners when navigating, which means as you navigate between views you end up activating lots of listeners you no longer need, which could cause bugs or performance issues. Can you check this?
> > ```
> 
> Updated code based on this: https://firebase.google.com/docs/firestore/query-data/listen#detach_a_listener
> 
> > ```
> > * By making this a one-page webapp, we've lost the ability to deep-link to particular pages, e.g. to link directly to the traffic page for a particular project. Could you add fixing that as an issue? 
> > ```
> 
> yes, will open an issue on that.

public/data.js
```diff
@@ -0,0 +1,62 @@
...
+            projectCollection = projectName,
+            unsubscribe = mediadb
+                .collection(`/metrics/rapid_pro/${projectCollection}/`)
+                .where("datetime", ">", offsetString)
+                .onSnapshot(res => {
+                    // Update data every time it changes in firestore
+                    DataController.updateData(res, data);
+                    onChange(data);
+                });
+        if (!hasListener) {
```
as2388: I don't think this does what you're trying to do. If `hasListener` is `false` then you'll create a snapshot listener then immediately unsubscribe it. If you're trying to unsubscribe a previously created listener, that won't happen here because you're only unsubscribing the new one.

public/app.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    static detachSnapshotListener() {
+        let DOMstrings = UIController.getDOMstrings(),
+            hasListener = false,
+            nodeList = document.querySelectorAll(DOMstrings.dropdownItem);
+        if (nodeList) {
+            // Makes sure all the snapshot listeners of the projects' traffic data are detached
+            Array.prototype.forEach.call(nodeList, obj => {
+                let project = obj.innerText;
```
as2388: This reads the project names out of the DOM, which creates a backward dependency that we were trying to avoid by restructuring the project like this. I think it would be better to have the watch functions in DataController return the unsubscribe function, save them to a variable in this class, then execute those functions and remove them from the variable when detaching.

as2388 (changes requested):
Couple of comments on the strategy for unsubscribing listeners

public/app.js
DanielMwendwa: resolved

public/data.js
DanielMwendwa: noted

public/data.js
```diff
@@ -0,0 +1,75 @@
...
+            .collection(`/metrics/rapid_pro/${projectCollection}/`)
+            .where("datetime", ">", offsetString)
+            .onSnapshot(res => {
+                // Update data every time it changes in firestore
+                DataController.updateData(res, data);
+                onChange(data);
+            });
+    }
+
+    static registerSnapshotListener(snapshot) {
```
as2388: This isn't a snapshot, it's the unsubscribe function

as2388 (approved):
Thanks @DanielMwendwa, approved subject to comments.

Please ask @lukechurch for review once those comments are addressed.

public/data.js
DanielMwendwa: resolved: 053fe1e

lukechurch:
I'm happy to delegate this one to @as2388

as2388:
> I'm happy to delegate this one to @as2388

In that case @DanielMwendwa you're good to merge and deploy this now :)

DanielMwendwa:
> > I'm happy to delegate this one to @as2388
> 
> In that case @DanielMwendwa you're good to merge and deploy this now :)

Thank you @as2388

PR #125, opened 2020-01-07
 Improve Message Coding table headers 

### LINKED ISSUE #120: Improve Message Coding table headers
We're seeing people interpret the dashboard as showing the total number of messages received in each dataset, and the fraction of WS/NC being interpreted as representative of the analysis dataset. These interpretations, however, are incorrect because the dashboard actually shows the number of unique messages and the fraction of those unique messages.

We could start to resolve this by renaming the table headers to e.g. "Unique Texts" rather than "Messages" etc. We might also add a description underneath the table that explains those columns and what can and cannot be inferred from them.

DanielMwendwa:
closes #120

### DIFF
--- a/public/coda/coding_progress.html
+++ b/public/coda/coding_progress.html
@@ -60,8 +60,8 @@
             <thead>
                 <tr class="table-heading">
                     <th scope="col">Dataset</th>
-                    <th scope="col">Messages</th>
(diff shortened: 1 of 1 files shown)


DanielMwendwa:
updates on #154

PR #131, opened 2020-01-16
used relative urls

DanielMwendwa:


IsaackMwenda (approved):

IsaackMwenda:
The pr should be use relative urls for 3rd party libraries.

DanielMwendwa:
> The pr should be use relative urls for 3rd party libraries.

Thanks

PR #132, opened 2020-01-16
deletion and update of code structure

DanielMwendwa:


IsaackMwenda (changes requested):
@DanielMwendwa Since this includes three independent changes, please create 3 new prs for removing old code, directory re-structure, and temporary operator fix. That way it will be easier to review the first two prs and allow us to deploy the temporarily fix independently as a branch without merging it.

DanielMwendwa:
> @DanielMwendwa Since this includes three independent changes, please create 3 new prs for removing old code, directory re-structure, and temporary operator fix. That way it will be easier to review the first two prs and allow us to deploy the temporarily fix independently as a branch without merging it.

yeah. thought so as well

PR #143, opened 2020-01-22
deep link to particular pages 

### LINKED ISSUE #129: Deep-link to Particular Pages
By making OPs dashboard a one-page webapp, we've lost the ability to deep-link to particular pages, e.g. to link directly to the traffic page for a particular project.

DanielMwendwa:
Closes #129

### DIFF
--- a/public/app.js
+++ b/public/app.js
@@ -42,7 +42,7 @@ class Controller {
         if (e.target && e.target.nodeName == "A") {
             Controller.resetUI();
             DataController.detachSnapshotListener();
-            location.hash = "coding_progress";
+            window.location.hash = "coding_progress";
             Controller.displayCodingProgress();
         }
     }
@@ -53,7 +53,7 @@ class Controller {
             DataController.detachSnapshotListener();
             console.log(e.target.innerText);
             let project = e.target.innerText;
-            location.hash = `traffic-${project}`;
+            window.location.hash = `traffic-${project}`;
             Controller.displayProject(project);
         }
     }
@@ -64,11 +64,11 @@ class Controller {
         activeProjectsData.forEach(project => {
             activeProjects.push(project.project_name);
         });
-        if (activeProjects.includes(page_route)) {
-            let project = page_route;
+        let project = page_route.split("-")[1];
+        if (activeProjects.includes(project)) {
             Controller.displayProject(project);
         } else {
-            // update the URL and replace the item in the browser history 
+            // update the URL and replace the item in the browser history
             // without reloading the page
             history.replaceState(null, null, " ");
             Controller.displayCodingProgress();
@@ -89,8 +89,14 @@ class Controller {
         if (page_route) {
             if (page_route == "coding_progress") {
                 Controller.displayCodingProgress();
+            } else if (page_route.startsWith("traffic-")) {
+                DataController.watchActiveProjects(Controller.displayDeepLinkedTrafficPage);
+            } else {
+                // update the URL and replace the item in the browser history
+                // without reloading the page
+                history.replaceState(null, null, " ");
+                Controller.displayCodingProgress();
             }
-            DataController.watchActiveProjects(Controller.displayDeepLinkedTrafficPage);
         } else {
             Controller.displayCodingProgress();
         }
--- a/public/app.js
+++ b/public/app.js
@@ -53,7 +53,7 @@ class Controller {
             DataController.detachSnapshotListener();
             console.log(e.target.innerText);
             let project = e.target.innerText;
-            location.hash = project;
+            location.hash = `traffic-${project}`;
             Controller.displaySelectedProject(project);
         }
     }
--- a/public/app.js
+++ b/public/app.js
@@ -19,15 +19,31 @@ class Controller {
         document.querySelector(DOMstrings.graphContainer).innerHTML = "";
     }
 
+    static displayCodingProgress() {
+        // Add the coding progress section to the UI
+        UIController.addCodingProgressSection();
+        // Get data for coding progress table
+        let unsubscribeFunc = DataController.watchCodingProgress(UIController.updateProgressUI);
+        DataController.registerSnapshotListener(unsubscribeFunc);
+    }
+
+    static displaySelectedProject(project) {
+        // Add the graphs container to the UI
+        UIController.addGraphs(project);
+        // Update and show the Graphs
+        let unsubscribeFunc = DataController.watchProjectTrafficData(
+            project,
+            GraphController.updateGraphs
+        );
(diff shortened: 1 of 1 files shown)


public/app.js
```diff
@@ -19,15 +19,31 @@ class Controller {
         document.querySelector(DOMstrings.graphContainer).innerHTML = "";
     }
 
+    static displayCodingProgress() {
+        // Add the coding progress section to the UI
+        UIController.addCodingProgressSection();
+        // Get data for coding progress table
```
as2388, line 25: You're not actually getting data here, you're setting up a snapshot listener.

I don't think you actually need a comment here - it's pretty clear from the function names what's going on. (Let me know if you disagree)

public/app.js
```diff
@@ -19,15 +19,31 @@ class Controller {
...
+        let unsubscribeFunc = DataController.watchCodingProgress(UIController.updateProgressUI);
+        DataController.registerSnapshotListener(unsubscribeFunc);
+    }
+
+    static displaySelectedProject(project) {
```
```suggestion
    static displayProject(project) {
```
On reading the name displayProject, my expectation is that you'll be passing in the project name to be displayed.

On reading the name displaySelectedProject, my expectation is that the function will somehow already know which project is selected, presumably by a class or global variable somewhere, so this function will take 0 arguments.

What you've implemented is the right thing to do, so I think better to call it `displayProject`

public/app.js
```diff
@@ -56,11 +80,17 @@ class Controller {
         Controller.setupEventListeners();
         // Add the dropdown menu to the UI
         DataController.watchActiveProjects(UIController.addDropdownMenu);
-        // Add the coding progress section to the UI
-        UIController.addCodingProgressSection();
-        // Get data for coding progress table
-        let unsubscribeFunc = DataController.watchCodingProgress(UIController.updateProgressUI);
-        DataController.registerSnapshotListener(unsubscribeFunc);
+        // Get hash value
+        Controller.hash = window.location.hash.substring(1);
```
as2388: I find Controller.hash a confusing name, because I'd expect that to be the hash of the Controller object, not the hash component of the url. Suggest something like e.g. url_hash_component, or page_route

public/app.js
```diff
@@ -37,14 +53,22 @@ class Controller {
...
-                project,
-                GraphController.updateGraphs
-            );
-            DataController.registerSnapshotListener(unsubscribeFunc);
+            location.hash = project;
+            Controller.displaySelectedProject(project);
+        }
+    }
+
+    static deepLinktoPage(activeProjectsData) {
```
as2388: This name is confusing, I can't tell whether you're setting a deep link or reading a deep link.

How about displayDeepLinkedTrafficPage? 

It might help readability to pass in the link's hash component as a variable too.

public/app.js
```diff
@@ -37,14 +53,22 @@ class Controller {
...
+        let activeProjects = [];
+        activeProjectsData.forEach(project => {
+            activeProjects.push(project.project_name);
+        });
+        if (activeProjects.includes(Controller.hash)) {
+            let project = Controller.hash;
+            Controller.displaySelectedProject(project);
+        } else {
+            history.replaceState(null, null, " ");
```
as2388, line 73: Could you add a brief comment justifying this?

public/app.js
```diff
@@ -37,14 +53,22 @@ class Controller {
...
             let project = e.target.innerText;
-            // Add the graphs container to the UI
-            UIController.addGraphs(project);
-            // Update and show the Graphs
-            let unsubscribeFunc = DataController.watchProjectTrafficData(
-                project,
-                GraphController.updateGraphs
-            );
-            DataController.registerSnapshotListener(unsubscribeFunc);
+            location.hash = project;
```
```suggestion
            location.hash = `traffic-${project}`;
```

as2388 (changes requested):
Thanks @DanielMwendwa.

What would be really nice is if these were formatted e.g. public/traffic/IOM or public/coding_progress etc., but that's not at all urgent. Happy to accept and deploy this as an incremental improvement in the meantime - requesting changes so I can take another look once the comments in the rest of the code have been addressed.

public/app.js
DanielMwendwa: The comment was there before I changed the snapshot listener function name. I agree Its better to remove the comment.

public/app.js
DanielMwendwa: Added

public/app.js
DanielMwendwa: noted

public/app.js
```diff
@@ -19,15 +19,31 @@ class Controller {
...
     static navigateToCodingProgress(e) {
         if (e.target && e.target.nodeName == "A") {
             Controller.resetUI();
             DataController.detachSnapshotListener();
-            // Add the coding progress section to the UI
-            UIController.addCodingProgressSection();
-            // Get data for coding progress table
-            let unsubscribeFunc = DataController.watchCodingProgress(UIController.updateProgressUI);
-            DataController.registerSnapshotListener(unsubscribeFunc);
+            location.hash = "coding_progress";
```
as2388: Can this be window.location.hash?

(You read from window.location.hash, and I think that's a clearer name)

public/app.js
```diff
@@ -56,11 +83,17 @@ class Controller {
...
-        let unsubscribeFunc = DataController.watchCodingProgress(UIController.updateProgressUI);
-        DataController.registerSnapshotListener(unsubscribeFunc);
+        // Get hash value
+        let page_route = window.location.hash.substring(1);
+        // Navigate appropriately according to the hash value
+        if (page_route) {
+            if (page_route == "coding_progress") {
+                Controller.displayCodingProgress();
+            }
+            DataController.watchActiveProjects(Controller.displayDeepLinkedTrafficPage);
```
as2388: This isn't safe, there's no guarantee that this is a valid traffic hash.

Consider:
```javascript
if (page_route == "coding_progress") {
    ...
} else if (page_route.startswith("traffic-") {
   ...
} else {
    ...
}
```
or use a switch.

as2388 (changes requested):
Sorry, this doesn't work for me. If I paste a hash link into my browser it just deletes it.

I think activeProjects.includes(page_route) isn't working

public/app.js
```diff
@@ -37,14 +53,25 @@ class Controller {
...
+        }
+    }
+
+    static displayDeepLinkedTrafficPage(activeProjectsData) {
+        let activeProjects = [],
+            page_route = window.location.hash.substring(1);
+        activeProjectsData.forEach(project => {
+            activeProjects.push(project.project_name);
+        });
+        let project = page_route.split("-")[1];
```
```suggestion
        let project = page_route.split("traffic-")[1];
```

as2388 (approved):

PR #156, opened 2020-02-13
Fix the last updated timestamp alerts showing false positives

### LINKED ISSUE #139: Last Updated alert shows false positives
The 'last updated' timestamp often shows in red unnecessarily. For example: 
image

Given that the sms service only runs for the previous complete ten-minute block since midnight (17:10-17:20) in this case and that it only runs once every 10 minutes, it could be until 17:30 until we get the block with the last update 17:10, so this shouldn't be red. I'd suggest waiting 25-30 minutes before reporting a problem.

### LINKED ISSUE #140: Last Update alert shows as red on tick but not on refresh
I've observed the 'last updated' timestamp showing in red, but then in black when I refresh the page, even when the date doesn't change.

Possibly related to #139.

as2388:
I think the false positives were being displayed when there was a genuine true positive missed update (which could happen while a machine was sleeping, and cause a UI updated on wake before the new data synced) and then no clearing of the alert once the new data had synced.

To fix this I had to:
 - Update the interval to clear the alert classes once we receive new data that's < 30 minutes old.
 - Clear the previous interval timer, otherwise it would keep firing, resulting in a "fight" between timers that were trying to set the alert classes and timers that were trying to clear it.

Fixes #139 and fixes #140.

### DIFF
--- a/public/app.js
+++ b/public/app.js
@@ -17,6 +17,7 @@ class Controller {
         let DOMstrings = UIController.getDOMstrings();
         document.querySelector(DOMstrings.codingProgressContainer).innerHTML = "";
         document.querySelector(DOMstrings.graphContainer).innerHTML = "";
+        GraphController.clearTimers();
     }
 
     static displayCodingProgress() {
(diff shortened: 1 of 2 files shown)


DanielMwendwa (approved):
Thanks @as2388

IsaackMwenda (approved):

PR #158, opened 2020-02-15
standardized mno colors

### LINKED ISSUE #138: dynamically change color scheme to be equal to the number of operators

### LINKED ISSUE #147: Standardize MNO colors
Let's standardize the colors between Somali and Kenyan MNos. Set Kenyan_telephone color to `#3DB44B` , NC to `#31CECE` and telegram to `#F032E6`

DanielMwendwa:
Closes: #147 and Closes: #138
The graphs use the custom color scheme that is used to generate the legends. I have updated the code to be able to alter the color scheme depending on the color we need to assign a given operator else we assign the operator a unique color. I think this reliable because if we assign each operator a color identity only (which will be hard coded)  we would not be able to assign identity color to a new operator. To ensure that our app doesn't break in case the list of operators is long we should enlarge the list of unique colors.


IsaackMwenda (commented):
Hey @DanielMwendwa adding an operator -> color string lookup table in firestore looks like a cleaner approach for this instead of hard coding colors to the front end. Let us do that then refactor this to use it.

DanielMwendwa:
@IsaackMwenda I've implemented a better approach where by I've given each operator a color after which I generate the color scheme. We will only update the `operators_indentity` object to acquire data from firestore once you add the table on this project's firestore.

public/graph.js
```diff
@@ -199,19 +199,17 @@ class GraphController {
...
+            operators_with_color_identity = Object.keys(MNOColors);
+
+        // Generate color scheme based on operators identity
+        operators.forEach((operator, index) => {
+            if (operators_with_color_identity.includes(operator)) {
+                color_scheme[index] = MNOColors[operator];
+            }
+        });
+
+        let color = d3.scaleOrdinal(color_scheme),
```
```suggestion
        let color = d3.scaleOrdinal(mno_color_scheme),
```

public/graph.js
```diff
@@ -199,19 +199,17 @@ class GraphController {
...
             colorReceived = d3.scaleOrdinal(color_scheme).domain(receivedKeys),
             colorSent = d3.scaleOrdinal(color_scheme).domain(sentKeys),
             colorFailed = d3.scaleOrdinal(["#a82e2e"]).domain(["total_errored"]);
```
IsaackMwenda: Have realized this is close enough with somtel we should probably use #ff0000
```suggestion
            colorFailed = d3.scaleOrdinal(["#ff0000"]).domain(["total_errored"]);
```

IsaackMwenda (approved):

PR #161, opened 2020-02-17
Updated the day view x-axis labels to help think about engagement cycles

### LINKED ISSUE #152: Update the day view x-axis labels to help think about engagement cycles
The labels used on the 1 day view of the traffic graphs make it a bit hard for me to read some of the information I want from them. They're great for reading the number of messages received today, and from each operator, but not for thinking about weekly engagement cycles.

In the first instance, I'd suggest adding the date to every day (not just every 5), and an initial for the day of the week. Will happily take other suggestions, especially if that approach turns out to be too cluttered.

DanielMwendwa:
engagement_cycles
Fixes: #152

### DIFF
--- a/public/graph.js
+++ b/public/graph.js
@@ -16,6 +16,7 @@ class GraphController {
         let isYLimitReceivedManuallySet = false,
             isYLimitSentManuallySet = false,
             dayDateFormat = d3.timeFormat("%Y-%m-%d"),
+            dayDateFormatWithWeekdayName = d3.timeFormat("%Y-%m-%d:%a"),
             operators = new Set();
 
         // Clear previous graphs before redrawing
@@ -526,8 +527,8 @@ class GraphController {
                 .call(
                     d3
                         .axisBottom(x)
-                        .ticks(d3.timeDay.every(4))
-                        .tickFormat(dayDateFormat)
+                        .ticks(d3.timeDay.every(1))
+                        .tickFormat(dayDateFormatWithWeekdayName)
                 )
                 // Rotate axis labels
                 .selectAll("text")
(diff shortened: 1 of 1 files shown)


public/graph.js
```diff
@@ -16,6 +16,7 @@ class GraphController {
         let isYLimitReceivedManuallySet = false,
             isYLimitSentManuallySet = false,
             dayDateFormat = d3.timeFormat("%Y-%m-%d"),
```
IsaackMwenda, line 18: Lets format this and use it instead of creating a new variable

public/graph.js
```diff
@@ -128,7 +129,7 @@ class GraphController {
             sentDataStackedDaily = stackSentDaily(dailySentTotal);
 
         //Create margins for the three graphs
-        const Margin = { top: 40, right: 100, bottom: 90, left: 70 },
+        const Margin = { top: 40, right: 100, bottom: 105, left: 70 },
```
IsaackMwenda: The bars seem more misaligned now, please retest and readjust.

IsaackMwenda (changes requested):
Requesting changes because of the misaligned bars.

public/graph.js
DanielMwendwa: ` data.forEach(function(d) {
            d.datetime = new Date(d.datetime);
            d.day = dayDateFormat(new Date(d.datetime));`
due to this if we format the first variable we'll get the error below
err

IsaackMwenda: Ahh i see thanks.

public/graph.js
```diff
@@ -789,13 +790,13 @@ class GraphController {
                 d3.select("#lastUpdated").classed("text-stale-info alert alert-stale-info", false);
             }
         }
-        
+
```
```suggestion
```

public/graph.js
```diff
@@ -789,13 +790,13 @@ class GraphController {
...
         if (GraphController.lastUpdateTimer) {
             clearInterval(GraphController.lastUpdateTimer);
         }
         GraphController.lastUpdateTimer = setInterval(setLastUpdatedAlert, 1000);
     }
-    
+
```
```suggestion
```

IsaackMwenda (approved):
Thanks

PR #163, opened 2020-02-19
Fix last updated timestamp

DanielMwendwa:


IsaackMwenda (approved):
Let us chat separately on how we can make the coding progress and traffic graph timestamps consistent either UK or local time.

PR #165, opened 2020-02-21
Selected Tab to change to 'Traffic' when viewing a traffic page

### LINKED ISSUE #164: Selected Tab doesn't change to 'Traffic' when viewing a traffic page
When viewing a traffic page, the selected tab still shows as "message coding":
image
(notice the red line under the message coding tab).

Low priority, but it does make navigation slightly confusing.

DanielMwendwa:
Closes: #164

### DIFF
--- a/public/UI.js
+++ b/public/UI.js
@@ -5,9 +5,12 @@ class UIController {
             projectMenu: ".project-menu",
             codingProgressLinkSelector: ".coding-progress-link",
             codingProgressContainer: ".coding-progress-container",
+            trafficsLinkSelector: ".traffics-link",
             graphContainer: ".graph-container",
             logoutBtn: ".logout-btn",
-            dropdownItem: ".dropdown-item"
+            dropdownItem: ".dropdown-item",
+            activeLinkClassName: "active-link",
+            activeLinks: "a.active-link"
         };
     }
 
@@ -168,7 +171,7 @@ class UIController {
             </section> 
             <div class="card shadow total_failed_sms_graph my-4"></div> 
         </div> `,
-        // Insert the HTML into the DOM
+            // Insert the HTML into the DOM
             newHtml = html.replace("%collection%", title);
         document
             .querySelector(DOMstrings.codingProgressContainer)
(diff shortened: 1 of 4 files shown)


as2388 (changes requested):
This is small graphical glitch with the left edge of the selected tab line on the traffic tab:
image

Compare with message coding, which works:
image

as2388 (changes requested):
image
This still doesn't look quite right, because the separating border is now double-width and the red border isn't.

But I see the problem now - you need to be able to show the left border of the traffic tab to make the highlight work, without increasing the total width of the separating border. I suggest either removing the right-border from coding-progress tab when it isn't selected, or shifting the left margin of the traffic tab to compensate for the wider border.

Also, when I refresh the page when viewing traffic, the message coding tab shows as selected for a few seconds before switching over to the traffic tab. Given that the URL is known at page load, can you set the correct tab immediately rather than a few seconds later?

DanielMwendwa:
> image
> This still doesn't look quite right, because the separating border is now double-width and the red border isn't.
> 
> But I see the problem now - you need to be able to show the left border of the traffic tab to make the highlight work, without increasing the total width of the separating border. I suggest either removing the right-border from coding-progress tab when it isn't selected, or shifting the left margin of the traffic tab to compensate for the wider border.
> 
> Also, when I refresh the page when viewing traffic, the message coding tab shows as selected for a few seconds before switching over to the traffic tab. Given that the URL is known at page load, can you set the correct tab immediately rather than a few seconds later?

Thanks for the idea

as2388 (changes requested):
Looks much better :)

There's a slight twitch on load due to the tabs being 2px smaller in height then when one of them has a border. Can you add an appropriate margin to fix that too?

(There's also the issue that they twitch while the logout icon is loading, but that's a separate issue)

as2388 (changes requested):
Sorry to go round on this again, but using line-height moves the text of the selected tab up by 2px, which is probably not what you want:
image

If you can reduce the line-height of selected items, I think that should work. 
(CSS is such a pain sometimes, I know).

as2388 (approved):
This looks really good, thank you @DanielMwendwa.

Please check whether the code freeze is still in effect before merging.

PR #168, opened 2020-03-09
add 10 minutes to lastupdated timestamp

### LINKED ISSUE #142: Clarify last updated timestamp
The current timestamp is possibly misleading, because it shows the start time of the last block of sms stat updates e.g. if the last update is for the range 12:40-12:50 then the last update will report as 12:40. I think it would better to report the last update as 12:50 instead (i.e. the end time for the block of updates).

DanielMwendwa:
Closes #142

### DIFF
--- a/public/graph.js
+++ b/public/graph.js
@@ -787,7 +787,7 @@ class GraphController {
             let currentTime = new Date(),
                 difference_ms = (currentTime.getTime() - lastUpdateTimeStamp.getTime()) / 60000,
                 difference_minutes = Math.floor(difference_ms % 60);
-            if (difference_minutes > 30) {
+            if (difference_minutes > 20) {
                 d3.select("#lastUpdated").classed("text-stale-info alert alert-stale-info", true);
             } else {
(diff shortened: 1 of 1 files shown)


public/graph.js
```diff
@@ -769,12 +769,15 @@ class GraphController {
...
         // Update timestamp of update and reset formatting
-        const lastUpdateTimeStamp = new Date(
+        let lastUpdateTimeStamp = new Date(
             Math.max.apply(
                 null,
                 data.map(d => new Date(d.datetime))
             )
         );
+        lastUpdateTimeStamp.setMinutes(lastUpdateTimeStamp.getMinutes() + 10);
+        lastUpdateTimeStamp = new Date(lastUpdateTimeStamp);
```
as2388, line 779: I think you need to update the warning threshold down by 10 minutes in setLastUpdatedAlert() (to 20 minutes instead of 30?)

as2388 (approved):
Thanks @DanielMwendwa.

Approved if you update the last update threshold accordingly.

public/graph.js
DanielMwendwa: Updated

PR #170, opened 2020-03-13
Modularize Graph js

### LINKED ISSUE #130: Break down graph update function into different modules
The update function for plotting the graphs is too long and it makes it hard to understand and modify the code. I suggest defining the functions inside the update function inside the class and just invoking them in the update function.

DanielMwendwa:
Closes #130 :  Separating the functionality of the program into independent, interchangeable modules, such that each contains everything necessary to execute only one aspect of the desired functionality, i.e., Separated logic of data manipulation from that of plotting graphs

The main aim of this is to break down the update function into sub modules to help keep track of the process happening in the update function and also to enable making ease of making changes since the parts concerned with manipulating data are separated from the logic of plotting the graphs.
Things to note: variables defined in static method are not accessible to other static methods that's why I'm using a lot of static properties to make them accessible every where in the class. variables not declared as static properties are only available in the scope of the static method.
I've renamed the class name since it is used almost everywhere in the code to improve readability of the code.

### DIFF
--- a/public/app.js
+++ b/public/app.js
@@ -17,7 +17,7 @@ class Controller {
         let DOMstrings = UIController.getDOMstrings();
         document.querySelector(DOMstrings.codingProgressContainer).innerHTML = "";
         document.querySelector(DOMstrings.graphContainer).innerHTML = "";
-        GraphController.clearTimers();
+        GC.clearTimers();
     }
 
     static displayCodingProgress() {
@@ -34,7 +34,7 @@ class Controller {
         // Update and show the Graphs
         let unsubscribeFunc = DataController.watchProjectTrafficData(
             project,
-            GraphController.updateGraphs
+            GC.updateGraphs
         );
         DataController.registerSnapshotListener(unsubscribeFunc);
     }
--- a/public/graph.js
+++ b/public/graph.js
@@ -402,93 +402,7 @@ class GC {
             d3.select("#buttonYLimitReceived").property("value", yLimitReceived);
             d3.select("#buttonYLimitSent").property("value", yLimitSent);
             GC.drawOneDayReceivedGraph(data, yLimitReceived);
-            drawOneDaySentGraph(yLimitSent);
-        }
-
-        function drawOneDaySentGraph(yLimitSent) {
-            // Set Y axis limit to max of daily values or to the value inputted by the user
-            let yLimitSentTotal = d3.max(GC.dailySentTotal, d => d.total_sent);
-
-            if (GC.isYLimitSentManuallySet != true) {
-                yLimitSent = yLimitSentTotal;
-            }
-
-            GC.xMin = d3.min(data, d => new Date(d.day));
-            GC.xMax = d3.max(data, d => GC.addOneDayToDate(d.day));
-            // set scale domains
-            GC.x.domain([GC.xMin, GC.xMax]);
-            GC.y_total_sent_sms.domain([0, yLimitSent]);
-
-            d3.selectAll(".redrawElementSent").remove();
-            d3.selectAll("#sentStack10min").remove();
-            d3.selectAll("#sentStack1day").remove();
-
-            // Add the Y Axis for the total sent sms graph
-            GC.total_sent_sms_graph
-                .append("g")
-                .attr("class", "axisSteelBlue")
-                .attr("class", "redrawElementSent")
-                .call(d3.axisLeft(GC.y_total_sent_sms));
-
-            // Create stacks
-            GC.sentLayer = GC.total_sent_sms_graph
-                .selectAll("#sentStack1day")
-                .data(GC.sentDataStackedDaily)
-                .enter()
-                .append("g")
-                .attr("id", "sentStack1day")
-                .attr("class", (d, i) => GC.sentKeys[i])
-                .style("fill", (d, i) => GC.color(i));
-
-            GC.sentLayer
-                .selectAll("rect")
-                .data(d => d)
-                .enter()
-                .append("rect")
-                .attr("x", d => GC.x(new Date(d.data.day)))
-                .attr("y", d => GC.y_total_sent_sms(d[1]))
-                .attr("height", d => GC.y_total_sent_sms(d[0]) - GC.y_total_sent_sms(d[1]))
-                .attr("width", GC.Width / Object.keys(GC.dailySentTotal).length);
-
-            //Add the X Axis for the total sent sms graph
-            GC.total_sent_sms_graph
-                .append("g")
-                .attr("class", "redrawElementSent")
-                .attr("transform", "translate(0," + GC.Height + ")")
-                .call(
-                    d3
-                        .axisBottom(GC.x)
-                        .ticks(d3.timeDay.every(1))
-                        .tickFormat(GC.dayDateFormatWithWeekdayName)
-                )
-                // Rotate axis labels
-                .selectAll("text")
-                .style("text-anchor", "end")
-                .attr("dx", "-.8em")
-                .attr("dy", ".15em")
-                .attr("transform", "rotate(-65)");
-
-            // Add X axis label for the total sent sms graph
-            GC.total_sent_sms_graph
-                .append("text")
-                .attr("class", "redrawElementSent")
-                .attr(
-                    "transform",
-                    "translate(" + GC.Width / 2 + " ," + (GC.Height + GC.Margin.top + 65) + ")"
-                )
-                .style("text-anchor", "middle")
-                .text("Date (Y-M-D)");
-
-            // Total Sms(s) graph title
-            GC.total_sent_sms_graph
-                .append("text")
-                .attr("class", "redrawElementSent")
-                .attr("x", GC.Width / 2)
-                .attr("y", 0 - GC.Margin.top / 2)
-                .attr("text-anchor", "middle")
-                .style("font-size", "20px")
-                .style("text-decoration", "bold")
-                .text("Total Outgoing Message(s) / day");
+            GC.drawOneDaySentGraph(data, yLimitSent);
         }
 
         // Update chart time unit on user selection
@@ -519,7 +433,7 @@ class GC {
             GC.isYLimitSentManuallySet = true;
             if (GC.chartTimeUnit == "1day") {
                 yLimitSent = this.value;
-                drawOneDaySentGraph(yLimitSent);
+                GC.drawOneDaySentGraph(data, yLimitSent);
             } else if (GC.chartTimeUnit == "10min") {
                 yLimitSentFiltered = this.value;
                 GC.draw10MinSentGraph(yLimitSentFiltered);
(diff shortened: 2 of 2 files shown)


DanielMwendwa (commented):
Renamed the class to an abbreviation since it will be used in the most part of the code

public/graph.js
```diff
@@ -1,250 +1,186 @@
...
+                    groupedData[`${operator}_${messageDirection}`] = d3.sum(
                         v,
-                        d => d[`${operator}_received`]
+                        d => d[`${operator}_${messageDirection}`]
                     );
                 });
-                receivedData["total_received"] = d3.sum(v, d => d.total_received);
-                return receivedData;
+                groupedData[`total_${messageDirection}`] = d3.sum(
+                    v,
```
IsaackMwenda: The formatting here is a bit odd. One line for this is would be great. Check if your IDE is causing this.

public/graph.js
```diff
@@ -254,554 +190,651 @@ class GraphController {
...
+            .selectAll("rect")
+            .data(d => d)
+            .attr("x", d => GC.x(d.data.datetime))
+            .attr("y", d => GC.y_total_sent_sms_range(d[1]))
+            .attr("height", d => GC.y_total_sent_sms_range(d[0]) - GC.y_total_sent_sms_range(d[1]))
+            .attr("width", GC.Width / Object.keys(data).length);
+    }
+
+    static setLastUpdatedAlert() {
+        // Calculate time diff bw current and lastUpdateTimeStamp
```
```suggestion
        // Calculate time diff between current and lastUpdateTimeStamp
```

public/graph.js
```diff
@@ -254,554 +190,651 @@ class GraphController {
...
+                d => GC.y_total_received_sms_range(d[0]) - GC.y_total_received_sms_range(d[1])
+            )
+            .attr("width", GC.Width / Object.keys(dataFilteredWeek).length);
+
+        //Add the X Axis for the total received sms graph
+        GC.total_received_sms_graph
+            .append("g")
+            .attr("class", "redrawElementReceived")
+            .attr("transform", "translate(0," + GC.Height + ")")
+            .call(
```
IsaackMwenda: fix formatting

public/graph.js
```diff
@@ -254,554 +190,651 @@ class GraphController {
...
-                .attr("x", Width / 2)
-                .attr("y", 0 - Margin.top / 2)
-                .attr("text-anchor", "middle")
-                .style("font-size", "20px")
-                .style("text-decoration", "bold")
-                .text("Total Outgoing Message(s) / day");
+        // Set default y-axis limits
+        let dataFilteredWeek = GC.data.filter(a => a.datetime > offsetWeek),
+            dataFilteredMonth = GC.data.filter(a => a.datetime > offsetMonth);
+
```
IsaackMwenda: add a comment for the block below

IsaackMwenda (approved):
This LGTM please request a secondary review from @as2388

public/app.js
```diff
@@ -48,7 +49,7 @@ class Controller {
         // Update and show the Graphs
         let unsubscribeFunc = DataController.watchProjectTrafficData(
             project,
-            GraphController.updateGraphs
+            GC.updateGraphs
```
IsaackMwenda, line 52: I still prefer the full name GraphController unless the abbreviation is more readable or common as it well explains its purpose to a new engineer. But will leave it to you to decide

public/graph.js
DanielMwendwa: Thanks!

public/app.js
DanielMwendwa: The issue is on graph.js class, i.e check line 110 - 119

as2388: Yep, I read GC as "garbage collector". Please leave as GraphController

public/graph.js
```diff
@@ -1,251 +1,181 @@
...
-        let isYLimitReceivedManuallySet = false,
-            isYLimitSentManuallySet = false,
-            dayDateFormat = d3.timeFormat("%Y-%m-%d"),
-            dayDateFormatWithWeekdayName = d3.timeFormat("%Y-%m-%d:%a"),
-            operators = new Set();
-
-        // Clear previous graphs before redrawing
-        d3.selectAll("svg").remove();
+    static setProperties() {
+        GC.TIMEFRAME_WEEK = 7;
```
as2388: Can you separate constants from user configurable options, and mark the constants in upper case. e.g. date formats are definitely constants, Ylimit manually set/received is not.

public/graph.js
```diff
@@ -1,251 +1,181 @@
...
-        offsetWeek.setDate(offsetWeek.getDate() - TIMEFRAME_WEEK);
-        offsetMonth.setDate(offsetMonth.getDate() - TIMEFRAME_MONTH);
-
-        // Set default y-axis limits
-        let dataFilteredWeek = data.filter(a => a.datetime > offsetWeek),
-            dataFilteredMonth = data.filter(a => a.datetime > offsetMonth);
-
-        // Group received data by day
-        let dailyReceivedTotal = d3
+    static GroupDataByDay(dataFilteredMonth, messageDirection) {
```
as2388:
```suggestion
    static groupDataByDay(dataFilteredMonth, messageDirection) {
```

public/graph.js
```diff
@@ -1,251 +1,181 @@
...
-        // let chartTimeUnit = "10min",
-        let isYLimitReceivedManuallySet = false,
-            isYLimitSentManuallySet = false,
-            dayDateFormat = d3.timeFormat("%Y-%m-%d"),
-            dayDateFormatWithWeekdayName = d3.timeFormat("%Y-%m-%d:%a"),
-            operators = new Set();
-
-        // Clear previous graphs before redrawing
-        d3.selectAll("svg").remove();
+    static setProperties() {
```
as2388, line 9: Is there a way of setting these without needing to use a magic setProperties() function call?

If yes, e.g. by pulling them into class constants, please do that.
If no, suggest converting GraphController to be non-static, and taking a reference to the DOM to update and all of these configuration objects in its constructor.

as2388: (class constants would look something like:
class GraphController {
    TIMEFRAME_WEEK: 7
}
but I don't write enough js these days to be sure of the exact syntax

public/graph.js
```diff
@@ -1,251 +1,181 @@
...
             d.total_sent = +d.total_sent;
             d.total_pending = +d.total_pending;
             d.total_errored = +d.total_errored;
             Object.keys(d.operators)
                 .sort()
                 .forEach(operator => {
-                    if (!(operator in operators)) {
-                        operators.add(operator);
+                    if (!(operator in GC.operators)) {
+                        GC.operators.add(operator);
```
as2388: console.warn() about this unexpected operator?

public/graph.js
```diff
@@ -1,251 +1,181 @@
...
+        GC.operators = new Set();
+    }
 
-        // format the data
+    static formatData(data) {
         data.forEach(function(d) {
             d.datetime = new Date(d.datetime);
-            d.day = dayDateFormat(new Date(d.datetime));
+            d.day = GC.dayDateFormat(new Date(d.datetime));
             d.total_received = +d.total_received;
```
as2388, line 24: What does this do? If it's just a type coercion to number, this shouldn't be necessary because the data is stored as numbers in the firestore. Please correct me if I'm missing something :)

public/graph.js
```diff
@@ -1,251 +1,181 @@
...
+                let groupedData = {};
+                GC.operators.forEach(operator => {
+                    groupedData[`${operator}_${messageDirection}`] = d3.sum(v,
+                        d => d[`${operator}_${messageDirection}`]
                     );
                 });
-                receivedData["total_received"] = d3.sum(v, d => d.total_received);
-                return receivedData;
+                groupedData[`total_${messageDirection}`] = d3.sum(v,d => d[`total_${messageDirection}`]
+                );
```
as2388: This can go on the previous line

public/graph.js
```diff
@@ -1,251 +1,181 @@
...
-                        d => d[`${operator}_received`]
+                let groupedData = {};
+                GC.operators.forEach(operator => {
+                    groupedData[`${operator}_${messageDirection}`] = d3.sum(v,
+                        d => d[`${operator}_${messageDirection}`]
                     );
                 });
-                receivedData["total_received"] = d3.sum(v, d => d.total_received);
-                return receivedData;
+                groupedData[`total_${messageDirection}`] = d3.sum(v,d => d[`total_${messageDirection}`]
```
as2388:
```suggestion
                groupedData[`total_${messageDirection}`] = d3.sum(v, d => d[`total_${messageDirection}`]
```

public/graph.js
as2388: What does dataFilteredMonth mean? Is this data required to be exactly one month?

public/graph.js
```diff
@@ -1,251 +1,181 @@
...
-                });
-                sentData["total_sent"] = d3.sum(v, d => d.total_sent);
-                return sentData;
-            })
-            .entries(dataFilteredMonth);
-
-        // Flatten nested data for stacking
-        for (let entry in dailySentTotal) {
-            let valueList = dailySentTotal[entry].value;
+    static FlattenNestedDataforStacking(messageDirection) {
```
as2388:
```suggestion
    static flattenNestedDataforStacking(messageDirection) {
```
because this is a method

public/graph.js
```diff
@@ -1,251 +1,181 @@
...
+        let nestedData;
+        if (messageDirection == "received") {
+            nestedData = GC.dailyReceivedTotal;
```
as2388: Where does this get set?

as2388 (commented):
Thanks @DanielMwendwa. I started reviewing this line-by-line but then realised most of the code was just being moved around slightly and wasn't very new. Line-by-line was also time-consuming.

Due to that, the start of the PR is more comment heavy than the end. Some general themes that came up:
 - Make sure your function names are formatted in camelCase.
 - Try to make your functions stateless units where possible, referring only to constants. At the moment you're accessing a lot of global state (GC.xxx) from functions, so you're relying on that state to have been set correctly by previous functions, and it's not immediately obvious when looking at a function which of the previous ones you needed to call were. It would be cleaner to pass that in in arguments where you can because then the control flow is more explicit.
 - Try to make functions as generic sounding as you can - there are some cases where the expected use is confusing the definitions e.g. groupDataByDay should be able to take any data that contains timestamped message counts, it shouldn't matter that they've been filtered by month, that's just how you'll call it in the typical case (unless I'm misunderstanding).

What might be really nice, although up to you whether to do in this PR or in future, is to turn GraphController into a class that can be initialised with some data and a DOM element, then you create 3 of those, one for each graph.

Can you take another pass with the above points in mind then request another review?

DanielMwendwa:
I'm working on the issues with high impact on performance then I'll resume updating this PR

DanielMwendwa:
> Thanks @DanielMwendwa. I started reviewing this line-by-line but then realised most of the code was just being moved around slightly and wasn't very new. Line-by-line was also time-consuming.
> 
> Due to that, the start of the PR is more comment heavy than the end. Some general themes that came up:
> 
> * Make sure your function names are formatted in camelCase.
> * Try to make your functions stateless units where possible, referring only to constants. At the moment you're accessing a lot of global state (GC.xxx) from functions, so you're relying on that state to have been set correctly by previous functions, and it's not immediately obvious when looking at a function which of the previous ones you needed to call were. It would be cleaner to pass that in in arguments where you can because then the control flow is more explicit.
> * Try to make functions as generic sounding as you can - there are some cases where the expected use is confusing the definitions e.g. groupDataByDay should be able to take any data that contains timestamped message counts, it shouldn't matter that they've been filtered by month, that's just how you'll call it in the typical case (unless I'm misunderstanding).
> 
> What might be really nice, although up to you whether to do in this PR or in future, is to turn GraphController into a class that can be initialised with some data and a DOM element, then you create 3 of those, one for each graph.
> 
> Can you take another pass with the above points in mind then request another review?

I have been working on your suggestion of turning GraphController into a class that can be initialised with some data and a DOM element. I've also come through the sentiments below:
"Mike Bostock, the creator of D3.js, recommends a pattern for creating reusable charts, implementing charts as closures with getter-setter methods. While adding some complexity to the chart implementation, setting configuration options becomes very straightforward for the caller by simply using method chaining". I'm analysing this pattern to see if it can help resolve this issue.

as2388:
> "Mike Bostock, the creator of D3.js, recommends a pattern for creating reusable charts, implementing charts as closures with getter-setter methods. While adding some complexity to the chart implementation, setting configuration options becomes very straightforward for the caller by simply using method chaining". I'm analysing this pattern to see if it can help resolve this issue.

One thing to note, before you go too deep into this approach, is that this is quite an old quote, so it may be worth checking if this is still a recommended way to do this. For example, some of the reasons to use the closure pattern, like javascript's insane this-binding and difficulty in creating objects, have now been fixed with classes and arrow functions, which didn't exist or weren't widely supported when that quote was made.

DanielMwendwa:
> > "Mike Bostock, the creator of D3.js, recommends a pattern for creating reusable charts, implementing charts as closures with getter-setter methods. While adding some complexity to the chart implementation, setting configuration options becomes very straightforward for the caller by simply using method chaining". I'm analysing this pattern to see if it can help resolve this issue.
> 
> One thing to note, before you go too deep into this approach, is that this is quite an old quote, so it may be worth checking if this is still a recommended way to do this. For example, some of the reasons to use the closure pattern, like javascript's insane this-binding and difficulty in creating objects, have now been fixed with classes and arrow functions, which didn't exist or weren't widely supported when that quote was made.

@as2388 please take a look at #238

PR #178, opened 2020-03-27
Remove unused code

DanielMwendwa:


DanielMwendwa:
Updated at #192  since this branch is dependent on another branch which has unresolved issues.

PR #182, opened 2020-04-01
Enable Tooltips for received one day view stacked bar graphs

DanielMwendwa:
Added tooltips for the received one day stacked bar graph. Look at this before I apply the tooltips to all the stacked bar graphs
Implementation: When one hovers on a bar graph I'll get the color of the bar and I'll use a function to look up the name of the operator that matches the color as defined in the variable `operator_identity`. I'll query the data for the given operator to output the number of messages on the tooltip.
tooltips

### DIFF
--- a/public/graph.js
+++ b/public/graph.js
@@ -467,10 +467,12 @@ class GraphController {
                         .html(d => { 
                             let receivedMessages = d.data[operatorNameWithMessageDirection],
                                 totalReceivedMessages = d.data.total_received,
-                                // Tooltip with operator name, date, no. of msg(s) & msg percentage in that day.
-                                tooltipContent = `<div>${operatorName.charAt(0).toUpperCase() + operatorName.slice(1)}</div>`;
-                            return tooltipContent += `<div>${receivedMessages} Message${receivedMessages !== 1 ? 's': ''} 
-                            (${Math.round((receivedMessages/totalReceivedMessages)*100)}%)</div>`;
+                                // Tooltip with operator name, no. of msg(s) & msg percentage in that day.
+                                tooltipContent = `<div>${receivedMessages} 
+                                (${Math.round((receivedMessages/totalReceivedMessages)*100)}%)
+                                ${operatorName.charAt(0).toUpperCase() + operatorName.slice(1)} 
+                                Message${receivedMessages !== 1 ? 's': ''} </div>`;
+                            return tooltipContent;
                     })
                     total_received_sms_graph.call(tip)
                     tip.show(d, n[i]).style("color", operatorColor)
--- a/public/graph.js
+++ b/public/graph.js
@@ -548,7 +548,8 @@ class GraphController {
                             let receivedMessages = d.data[operatorReceived],
                                 totalReceivedMessages = d.data.total_received,
                                 receivedDay = d.data.day,
-                                tooltipContent = `<div>${operatorName} ${dayDateFormatWithoutYear(new Date(receivedDay))}</div>`;
+                                tooltipContent = `<div>${operatorName.charAt(0).toUpperCase() + operatorName.slice(1)} 
+                                ${dayDateFormatWithoutYear(new Date(receivedDay))}</div>`;
                             return tooltipContent += `<div>${receivedMessages} Message${receivedMessages !== 1 ? 's': ''} 
                             (${Math.round((receivedMessages/totalReceivedMessages)*100)}%)</div>`;
                     })
--- a/public/graph.js
+++ b/public/graph.js
@@ -456,16 +456,16 @@ class GraphController {
                 .selectAll("rect")
                 .on("mouseover", (d, i, n) => {
                     // Get key of stacked data from the selection
-                    let operatorReceived = d3.select(n[i].parentNode).datum().key,
+                    let operatorNameWithMessageDirection = d3.select(n[i].parentNode).datum().key,
                         // Get operator name from the key
-                        operatorName = operatorReceived.replace('_received',''),
+                        operatorName = operatorNameWithMessageDirection.replace('_received',''),
                         // Get color of hovered rect
                         operatorColor = d3.select(n[i]).style("fill");
                     tip = d3.tip()
                         .attr("class", "tooltip")
                         .attr("id", "tooltip")
                         .html(d => { 
-                            let receivedMessages = d.data[operatorReceived],
+                            let receivedMessages = d.data[operatorNameWithMessageDirection],
                                 totalReceivedMessages = d.data.total_received,
                                 // Tooltip with operator name, date, no. of msg(s) & msg percentage in that day.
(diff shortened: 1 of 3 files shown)


public/graph.js
```diff
@@ -383,6 +383,50 @@ class GraphController {
...
+        function componentToHex(c) {
+            var hex = c.toString(16);
+                return hex.length == 1 ? "0" + hex : hex;
+            }
+
+        function rgbToHex(r, g, b) {
+            return "#" + componentToHex(r) + componentToHex(g) + componentToHex(b);
+        }
+
+        let operators_identity = {
```
IsaackMwenda: let us use the mno_colors table available in avf-dashboards firebase.

IsaackMwenda (changes requested):

public/graph.js
```diff
@@ -499,6 +530,36 @@ class GraphController {
...
+                .selectAll("rect")
+                .on("mouseover", (d, i, n) => {
+                    // Get color of hovered rect
+                    let rgb = d3.select(n[i]).style("fill")
+                    // Get color components from an rgb string
+                    rgb = rgb.substring(4, rgb.length-1).replace(/ /g, '').split(',');
+                    // Cast strings in array to int
+                    rgb = rgb.map((x) =>parseInt(x));
+                    let hex = rgbToHex(...rgb)
+                    let op = legendColorToOperator(hex)
```
IsaackMwenda: Change to a descriptive variable name, please.

public/styles.css
```diff
@@ -197,3 +197,12 @@ table > tbody > tr > td {
 .alert-stale-info {
     background-color: #f08080;
 }
+.tooltip2 {
```
IsaackMwenda:
```suggestion
.tooltip{	
```
?

public/graph.js
```diff
@@ -499,6 +530,36 @@ class GraphController {
...
+                
+                    tip = d3.tip()
+                        .attr("class", "tooltip2")
+                        .attr("id", "tooltip2")
+                        .html(d => {
+                            let num = d.data[`${op}_received`]
+                            let content = `<div>${op} ${num}</div>` 
+                            return content;
+                    })
+
```
```suggestion
```
remove unnecessary white spaces here and in the rest of the code

```suggestion
```

public/graph.js
```diff
@@ -499,6 +530,36 @@ class GraphController {
...
+                    // Cast strings in array to int
+                    rgb = rgb.map((x) =>parseInt(x));
+                    let hex = rgbToHex(...rgb)
+                    let op = legendColorToOperator(hex)
+                
+                    tip = d3.tip()
+                        .attr("class", "tooltip2")
+                        .attr("id", "tooltip2")
+                        .html(d => {
+                            let num = d.data[`${op}_received`]
```
```suggestion
                            let total_receved_no = d.data[`${op}_received`]
```
? or better

public/graph.js
```diff
@@ -499,6 +530,36 @@ class GraphController {
...
+                            let content = `<div>${op} ${num}</div>`
```
```suggestion
                            let tooltip_content = `<div>${op} ${num}</div>`  
```
?

public/graph.js
```diff
@@ -499,6 +530,36 @@ class GraphController {
...
+            receivedLayer
+                .selectAll("rect")
+                .on("mouseover", (d, i, n) => {
+                    // Get color of hovered rect
+                    let rgb = d3.select(n[i]).style("fill")
+                    // Get color components from an rgb string
+                    rgb = rgb.substring(4, rgb.length-1).replace(/ /g, '').split(',');
+                    // Cast strings in array to int
+                    rgb = rgb.map((x) =>parseInt(x));
+                    let hex = rgbToHex(...rgb)
```
IsaackMwenda: If operator colors are already stored in hex format, we should probably pull this from the table instead?

IsaackMwenda (changes requested):
Requesting changes because of a question re-color conversion

DanielMwendwa:
Issue: having trouble accessing the relevant values of the initial dataset for a mousover tooltip

Reason: `d` is the Array[2] at this point with the values defining baseline / topline, `d.data` is the complete original data object but is missing the information over which stack of the series I'm currently hovering.

Solution:  Get the name of the element accessing the parent's data, and using that name to get the value
https://github.com/AfricasVoices/OperationsDashboard/pull/182/commits/ba5229a175ffe228f529157a5b790d4e2ed98a13

public/graph.js
```diff
@@ -501,7 +502,38 @@ class GraphController {
...
+                        operatorName = operatorReceived.replace('_received',''),
+                        // Get color of hovered rect
+                        operatorColor = d3.select(n[i]).style("fill");
+                    tip = d3.tip()
+                        .attr("class", "tooltip")
+                        .attr("id", "tooltip")
+                        .html(d => { 
+                            let receivedMessages = d.data[operatorReceived],
+                                totalReceivedMessages = d.data.total_received,
+                                receivedDay = d.data.day,
```
IsaackMwenda: no need for the date info because it is already rep in the axes.

DanielMwendwa:
Updated tooltip content by removing date info

public/graph.js
```diff
@@ -450,7 +450,37 @@ class GraphController {
...
+                    let operatorReceived = d3.select(n[i].parentNode).datum().key,
+                        // Get operator name from the key
+                        operatorName = operatorReceived.replace('_received',''),
+                        // Get color of hovered rect
+                        operatorColor = d3.select(n[i]).style("fill");
+                    tip = d3.tip()
+                        .attr("class", "tooltip")
+                        .attr("id", "tooltip")
+                        .html(d => { 
+                            let receivedMessages = d.data[operatorReceived],
```
IsaackMwenda: operatorReceived?

public/graph.js
```diff
@@ -450,7 +450,37 @@ class GraphController {
...
+                                totalReceivedMessages = d.data.total_received,
+                                receivedDay = d.data.day,
```
IsaackMwenda: we no longer need this variable

public/graph.js
```diff
@@ -450,7 +450,37 @@ class GraphController {
...
+                                // Tooltip with operator name, date, no. of msg(s) & msg percentage in that day.
+                                tooltipContent = `<div>${operatorName.charAt(0).toUpperCase() + operatorName.slice(1)}</div>`;
```
IsaackMwenda: 28 (70%) Hormuud Messages?

IsaackMwenda (approved):
Great this LGTM! @DanielMwendwa. Its also going to be really useful. My only concern now is the duplication of the code once we apply this to outgoing and failed graphs. We might need to apply OOP for graph.js soon. An issue, for now, is okay. Thanks.

public/graph.js
DanielMwendwa: I'll update to `operatorNameWithMessageDirection` because it's the operator name concatenated with the msg direction

DanielMwendwa:
Thanks @IsaackMwenda

PR #183, opened 2020-04-01
undefined variables on master

DanielMwendwa:
declared variables using let and resolved the error


IsaackMwenda (approved):
Thanks for catching this.

PR #184, opened 2020-04-04
Fix ops-dashboard authentication

DanielMwendwa:
Fix ops-dashboard authentication by updating script src for updated sdks


IsaackMwenda (changes requested):
Thanks @DanielMwendwa redirect after login does not seem to work for me. Have you tested this?

IsaackMwenda (approved):

PR #186, opened 2020-04-06
update margin bottom of graphs to show x axis title

DanielMwendwa:


IsaackMwenda (approved):
The per 10 mins graph x-axis is incorrect please also fix that in a separate PR

PR #188, opened 2020-04-07
Fix yaxis for failed msg graph

DanielMwendwa:
Set the scale domain for failed graph when y max is greater than 0


IsaackMwenda (approved):
Thanks

PR #195, opened 2020-04-14
Handle listen errors

### LINKED ISSUE #185: Uncaught Error in onSnapshot: FirebaseError: Missing or insufficient permissions.
Screen Shot 2020-04-05 at 14 37 04

DanielMwendwa:
Closes: #185 
For more info on the importance of handling listen errors visit:
https://firebase.google.com/docs/firestore/query-data/listen#handle_listen_errors


IsaackMwenda (commented):
Hi @DanielMwendwa perhaps what would be more useful is a popup to the user that their authentication has been declined. console logging the error is a duplication of what is already being shown under there error tab
image

DanielMwendwa:
I have noticed this error comes from firestore since we have rules that check if the email is verified and if it comes from our domain. The front end only checks if we have a user object only.
This explains why a user can be sign in and not see any data and this error is logged.
I have updated the auth listener to check the domain of the email address. I don't expect this error to by pass the auth state listener where it is handled but incase it happens I've handled `permission denied error` on the first listener that access firestore for data to redirect to sign in page before any operations happen.

public/auth.js
```diff
@@ -2,7 +2,7 @@ class AuthController {
     // Authentication state listener
     static getUser() {
         firebase.auth().onAuthStateChanged(user => {
-            if (user) {
+            if (user && user.email.match(".*@africasvoices.org$")){
```
IsaackMwenda: We have users who have been authenticated in the past but not necessarily using avf email. This locks them out.

public/auth.js
```diff
@@ -17,4 +17,4 @@ class AuthController {
                 console.log(err);
             })
     }
-}
+}
```
IsaackMwenda, line 21: nit: add new line

public/data.js
```diff
@@ -27,19 +27,28 @@ class DataController {
             // Update data every time it changes in firestore
             DataController.updateData(res, activeProjects);
             onChange(activeProjects);
+        }, error => {
+            if (error.code == "permission-denied") {
+                // This error is only handled here because this is the second listener to be invoked 
+                // after Authentication state listener
+                alert(error.message) // Alert Missing or insufficient permissions.
```
IsaackMwenda: This doesn't seem to work, the user is only redirected to the auth page without knowing why their auth was not successful.

IsaackMwenda (changes requested):

public/auth.js
DanielMwendwa: Yeah I get it. The problem is that there are firestore rules that specify who can read from the database. The error `permission denied` comes when a user not using avf email logs in and he/she is denied to read from the db because of the rule below; for one to read the db the email should be avf's. That's why I have implemented the same on the app, since there is no reason to enable a person to login and cannot access data.
```
rules_version = "2";
service cloud.firestore {
  match /databases/{database}/documents {
    match /{document=**} {
      allow read: if request.auth.token.email_verified && request.auth.token.email.matches(".*@africasvoices.org$")
    }
  }
}
```

public/data.js
DanielMwendwa: I have tested its working but under some conditions.If you have not seen this alert message, it means you have been redirected from the code above that specifies that a user should be verified and using avf email. 
You will see this alert message when one is using avf email to login and the admin has restricted this person from accessing the db.

public/auth.js
DanielMwendwa: Please try to login with approved gmail address and you will see the issue above

IsaackMwenda: Am aware of the above rules. We have external collaborators who access the ops-dashboard using non-avf email address. Adding this check-in frontend locks them out from log-in whereas their authentication is handled in firebase. What we simply need here is a friendly frontend alert (not console logs) to users with details as to why their auth was declined.

DanielMwendwa: Updated authentication to give alert on denied permission

public/auth.js
IsaackMwenda: this is unresolved, you can also set up your editor to do this auto for you :)

public/data.js
```diff
@@ -27,19 +27,37 @@ class DataController {
             // Update data every time it changes in firestore
             DataController.updateData(res, activeProjects);
             onChange(activeProjects);
+        }, error => {
+            if (error.code == "permission-denied") {
+                firebase.auth().onAuthStateChanged(user => { 
+                    if (user) {
+                        name = user.displayName.split(" ").slice(0,-1)
```
IsaackMwenda: As much as this is nice to have let minimize the logic as this eventually takes a toll on the dashboard performance.

public/data.js
```diff
@@ -27,19 +27,37 @@ class DataController {
...
+                        // Error message - Alert Missing or insufficient permissions.
+                        alert(`${error.message} 
+                        \r Please ${name} use AVF email or liaise with Africa's Voices Foundation (AVF) to give you AVF email`) 
+                        if (user.email.match(".*@africasvoices.org$")) {
```
IsaackMwenda, line 37: This still locks out non -avf but authorized email addresses

IsaackMwenda (changes requested):

public/data.js
DanielMwendwa: Please note that the logic is after the error permission-denied (The user has been already been locked out at this point) and that the logic is used to generate feedback that would be relevant to the user.  The best thing to do for them to have access is to update firestore rules.

public/data.js
DanielMwendwa: Thanks, updated.

public/data.js
```diff
@@ -27,19 +27,36 @@ class DataController {
             // Update data every time it changes in firestore
             DataController.updateData(res, activeProjects);
             onChange(activeProjects);
+        }, error => {
+            if (error.code == "permission-denied") {
+                firebase.auth().onAuthStateChanged(user => { 
+                    if (user) {
+                        // Error message - Alert Missing or insufficient permissions.
+                        alert(`${error.message} 
+                        \r Please use AVF email or liaise with Africa's Voices Foundation (AVF) to give you AVF email`)
```
IsaackMwenda:
```suggestion
                        \r Please login using your AVF email`) 
```

public/data.js
```diff
@@ -27,19 +27,36 @@ class DataController {
...
+                console.log(error);
+            }     
         });
     }
 
     static watchCodingProgress(onChange) {
         return mediadb.doc("metrics/coda").onSnapshot(res => {
             onChange(res.data());
-        });
+        }, error => console.log(error));
```
IsaackMwenda, line 53: do we still need this?

IsaackMwenda (approved):

public/data.js
DanielMwendwa: Yes, after an error, the listener will not receive any more events, and there is no need to detach your listener

PR #197, opened 2020-04-15
Experiment D3 Transitions

DanielMwendwa:


DanielMwendwa:
I was able to create transition for the stacked bars but there was a problem with updates where by on update the transition started at the bottom of the graph instead of the heights of the bars, further more the graphs are redrawn and not updated from their previous state.

PR #198, opened 2020-04-15
Add tooltips for other stacked graphs

DanielMwendwa:
copied code from PR #182 and updated the variable according to the graph the tooltip is being applied and the message direction

### DIFF
--- a/public/graph.js
+++ b/public/graph.js
@@ -599,6 +599,37 @@ class GraphController {
                 .attr("height", d => y_total_sent_sms_range(d[0]) - y_total_sent_sms_range(d[1]))
                 .attr("width", Width / Object.keys(dataFilteredWeek).length);
 
+            // Add tooltip for the total sent sms graph
+            let tip;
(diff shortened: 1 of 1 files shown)


public/graph.js
```diff
@@ -361,6 +361,37 @@ class GraphController {
                 )
                 .attr("width", Width / Object.keys(dataFilteredWeek).length);
 
+            // Add tooltip for the total received sms graph
```
IsaackMwenda: We don't need tooltips for 10 mins view let us have it in per day only and add this if requested by the team.

IsaackMwenda (changes requested):

public/graph.js
DanielMwendwa: removed tooltips for 10 minute view

IsaackMwenda: Thanks

IsaackMwenda (approved):
Thanks, @DanielMwendwa!

Issue #115, opened 2019-11-14
Graphs sometimes render misaligned bars 


as2388:
See this example:
<img width="638" alt="image" src="https://user-images.githubusercontent.com/7963608/68877239-a9979e80-06fd-11ea-865e-47333972ea31.png">

I'm not sure what the cause is, but I think I've only observed this on pages that have been left open for a while, so suspect the redraw isn't working quite correctly.

DanielMwendwa:
Haven't noticed this yet on my side. Gathering information on this.

PR #201, opened 2020-04-16
Control x axis range

### LINKED ISSUE #40: Control x axis range
There are different scales that it's helpful to look at the graphs in. They are probably:

31 days
7 days
48 hours
24 hours

Let's add a drop down box that allows these to be selected.

DanielMwendwa:
Closes: #40, #115
Added a drop down menu to select the period you want to view the graphs in. The dropdown menu has 5 options:
1. Default - view the 10 min graph in 1 weeks timeframe and 1 day graph in 30 days timeframe
The other options change timeframe for both graphs:
2. 48 hours - to enable comparison of today and previous day number of msg
3. 7 days - to enable comparison of number msg received in a week
4. 14 days - to enable comparison no. of msg received between a week and its previous week
5. 30 days- to enable comparison of number msg received in a month

I would also like to mention that the bar charts show an extra day in each timeframe , for example, in a weeks timeframe it shows 8 days. That has always been there in the graphs. If concerned I will make the bar chart to be consistent with the time frame.

I have also opened #202 stemming from this PR to align the x axis ticks correctly to avoid confusion and also put some space between bars to avoid misalignment which is caused by the dependence of the width of the bars to the data used to plot the graphs. There is no reason to decouple the two because its helps in recalculation of the bars width when data changes and the misalignment happens in a few degrees.
timeframe


DanielMwendwa:
I have also noticed when you select a timeframe, it takes a while in the 10 min view to update, the problem is not even the time taken to update, it's that the graph changes quickly after taking a while and one can sometime not notice if the timeframe has changed. To enhance human interaction I suggest enabling transitions where by when a timeframe is selected the graphs fades out and the ones with the selected timeframe fade in, or something close to that to show the user that he/she is going to transition from one timeframe to another.

public/UI.js
```diff
@@ -208,10 +208,18 @@ class UIController {
                 <div>
                     <span class="txt-brown my-auto title"><b>%collection%</b></span>
                 </div>
-                <div>
+                <div class="d-md-flex">
```
IsaackMwenda, line 211: nit: add a lil bit of padding between the text and buttons.

public/graph.js
```diff
@@ -6,9 +6,10 @@ class GraphController {
         return newDate;
     }
 
-    static updateGraphs(data, projectName, MNOColors) {
-        const TIMEFRAME_WEEK = 7,
-            TIMEFRAME_MONTH = 30;
+    static updateGraphs(data, projectName, MNOColors, week=7, month=30) {
+        console.log(projectName)
```
IsaackMwenda: forgot this here? : )

public/graph.js
```diff
@@ -917,6 +918,16 @@ class GraphController {
             updateViewOneDay(yLimitReceived, yLimitSent, yLimitFailed);
         });
 
+        d3.select("#timeFrame").on("change", function() {
```
IsaackMwenda, line 947: The misalignment, especially on day->48 hours, is so visible I wouldn't want us to merge this then resolve in a diff branch so probably better to solve it here? 
So probably the first place to start this is changing from using the data to determine the bar width? especially on the 1-day view? Having a standard width and the use the data to set the bar heights.

IsaackMwenda (changes requested):

public/UI.js
DanielMwendwa: Ok

public/graph.js
DanielMwendwa: Thanks, updated

public/graph.js
DanielMwendwa: I have updated the graph bar width and x attributes and enabled little space between bars to avoid misalignment of bars as in issue  #115. 
Note: I have made the x axis ticks to be at the center of the bar graphs,I have noticed enabling them to be at the start of a bar chart makes it shift when the timeframe is switched altering the correctness of the graph

DanielMwendwa:
Enabled gridlines to enhance the graphs

public/graph.js
```diff
@@ -420,6 +426,25 @@ class GraphController {
             d3.selectAll(".redrawElementReceived").remove();
             d3.selectAll("#receivedStack10min").remove();
             d3.selectAll("#receivedStack").remove();
+            d3.selectAll(".grid").remove();
+
+            // Add the X gridlines
+            total_received_sms_graph.append("g")
```
IsaackMwenda: Lets split this in a separate pr, please.

public/graph.js
IsaackMwenda: let switch back to left-aligned ticks, and then fix the width mis-alignments. I suggest having fixed bar widths for each timeframe.

IsaackMwenda (changes requested):

public/graph.js
DanielMwendwa: I have noticed there is a problem with my implementation and there is no reason to calculate the x position and bar width. I will update them to default.

DanielMwendwa:
I have applied the above changes on received graph, the proceeding changes are the replication of the changes made on the received graph

public/graph.js
DanielMwendwa: Done

DanielMwendwa:
Thanks @IsaackMwenda I have updated the graphs as requested above.

DanielMwendwa:
Check Thursday 9th on week range graph there is 1 msg while on month range there is 200 msgs
`week range`
week range
`month range`
month range


I have made the changes above because I have realized the first bar chart in the graph for each timeframe is not correct. This is because the offset dates don't begin at the start of the day, some date offset actually start at 10pm of that day leaving the rest of the day msgs not to be included in the graph.

DanielMwendwa:
`1 week timeframe` shows 6 days time around midnight while through out the day 7 days
1 wk timeframe

This also resolves the issue of displaying less one bar chart at around midnight on graphs

public/graph.js
```diff
@@ -437,6 +440,7 @@ class GraphController {
                 .attr("class", (d, i) => receivedKeys[i])
                 .style("fill", (d, i) => color(i));
 
+            let innerPadding = 5
```
```suggestion
            let innerPadding = 2 
``` or 1.5 ?

File: public/graph.js [hunk 6]
```diff
@@ -448,7 +452,7 @@ class GraphController {
                     "height",
                     d => y_total_received_sms_range(d[0]) - y_total_received_sms_range(d[1])
                 )
-                .attr("width", Width / Object.keys(dailyReceivedTotal).length);
+                .attr("width", (Width / Object.keys(dailyReceivedTotal).length) - innerPadding);
```
IsaackMwenda: It would nice to have the padding at the start of the day rather than at the end. This will help us prevent the first bar from covering the y-axis bar. The alternative is we fix that separately and only reduce the current inner padding.
image

public/styles.css
```diff
@@ -206,3 +206,4 @@ table > tbody > tr > td {
     border-radius: 8px;			
     pointer-events: none;			
 }
+
```
IsaackMwenda:
```suggestion
```

IsaackMwenda (changes requested):
Thanks @DanielMwendwa  LGTM! only requesting changes so that I can take a look at the bar width changes before we merge.

public/graph.js
DanielMwendwa: Done

DanielMwendwa:
Resolved first bar of graph overlapping y axis path

public/graph.js
```diff
@@ -437,18 +440,20 @@ class GraphController {
                 .attr("class", (d, i) => receivedKeys[i])
                 .style("fill", (d, i) => color(i));
 
+            // Values to adjust x and width attributes
```
IsaackMwenda, line 446: add some details as to why we we need this

public/graph.js
```diff
@@ -50,6 +50,9 @@ class GraphController {
 
         offsetWeek.setDate(offsetWeek.getDate() - TIMEFRAME_WEEK);
         offsetMonth.setDate(offsetMonth.getDate() - TIMEFRAME_MONTH);
+        // Set date offsets to nearest midnight in the past
```
IsaackMwenda: add some more details to the comment

IsaackMwenda (approved):
Thanks for the patience @DanielMwendwa. We needed to get this right.

public/graph.js
DanielMwendwa: Done!

public/graph.js
DanielMwendwa: Done!

PR #202, opened 2020-04-17
Resolve rendering misaligned bars

### LINKED ISSUE #115: Graphs sometimes render misaligned bars

DanielMwendwa:
Closes: #115
View first #201 : I have only added one commit here to resolve the issue of rendering misaligned bars.


DanielMwendwa:
Changes appear in PR #201

PR #206, opened 2020-04-22
Add Gridlines to Graphs

DanielMwendwa:
Added grid lines to graphs as a visual aid & Changed how x-axis ticks are being generated from `.ticks(d3.timeDay.every(1)) ` to using date values from data used to plot a given graph. This avoids the error of x ticks from getting out of the graph width sometimes around midnight
x-axis-err

### DIFF
--- a/public/graph.js
+++ b/public/graph.js
@@ -526,7 +526,6 @@ class GraphController {
                 })
 
             // "Add the X Axis for the total received sms graph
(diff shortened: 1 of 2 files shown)


IsaackMwenda:
Thanks for picking that out. Lets break this or into two add-gridlines and x-axis-ticks-position-fix

PR #211, opened 2020-04-28
Compute coding progress slow to load

### LINKED ISSUE #150: Compute coding progress slow to load
It shouldn't take 10s to download 100-200 documents from Firestore and draw a table. There's probably an inefficiency somewhere in the javascript.

DanielMwendwa:
I've noticed we're preforming extensive DOM manipulation and also sorting inefficiently. I've tried a simplified implementation of the same taking into consideration all the inefficiency we had before. 
Considering the factors below I've been able to plot an efficient table for coding progress with D3:
-> D3 allows you to **bind arbitrary data to a Document Object Model (DOM)**, and then apply **data-driven transformations to the document**.
-> D3 solves the crux of the problem: **efficient manipulation of documents based on data**. This avoids proprietary representation and affords extraordinary flexibility, exposing the full capabilities of web standards such as HTML, SVG, and CSS. With minimal overhead, D3 is extremely fast, supporting large datasets and dynamic behaviors for interaction and animation.
-> **Modifying documents using the W3C DOM API is tedious**: the method names are verbose, and the imperative approach requires manual iteration and bookkeeping of temporary state.
-> D3 employs a **declarative approach**, operating on arbitrary sets of nodes called selections. 
Closes: #150


public/data.js
```diff
@@ -32,7 +32,27 @@ class DataController {
...
+                codingProgress["Unique Texts"] = messagesCount
+                codingProgress["Unique Texts with a label"] = messagesWithLabel 
+                codingProgress["Done"] =  ((100 * messagesWithLabel) / messagesCount).toFixed(2);
+                codingProgress["Wrong Scheme messages"] = wrongSchemeMessages != null ? wrongSchemeMessages : "-";
+                codingProgress["WS %"] = wrongSchemeMessages != null ? ((100 * wrongSchemeMessages) / messagesCount).toFixed(2): "-";
+                codingProgress["Not Coded messages"] = notCodedMessages != null ? notCodedMessages : "-";
+                codingProgress["NC %"] = notCodedMessages != null ? ((100 * notCodedMessages) / messagesCount).toFixed(2) : "-";
+                data.push(codingProgress)
+            }
+
```
IsaackMwenda:
```suggestion
```

public/index.html
```diff
@@ -53,6 +53,7 @@
     <script src="data.js"></script>
     <script src="UI.js"></script>
     <script src="graph.js"></script>
+    <script src="table.js"></script>
```
IsaackMwenda: codingprogress or codingprogresstable ?

public/data.js
```diff
@@ -32,7 +32,27 @@ class DataController {
 
     static watchCodingProgress(onChange) {
         return mediadb.doc("metrics/coda").onSnapshot(res => {
-            onChange(res.data());
+            let data = []
+            let codaData = res.data();
```
IsaackMwenda:
```suggestion
            let codingProgressData = res.data();
```

public/table.js
```diff
@@ -0,0 +1,106 @@
...
+        };
+    };
+
+    static stringCompare(a, b, order) {
+        if (order === "descending") 
+            return a.localeCompare(b, 'en', { sensitivity: 'base' });
+        return b.localeCompare(a, 'en', { sensitivity: 'base' });
+    };
+
+    static sortNumber(a,b, order) {
```
IsaackMwenda:
```suggestion
    static sortNumber(a, b, order) {
```

public/table.js
```diff
@@ -0,0 +1,106 @@
...
+    static jsonToArray(json) {
+        var arr = [];
+        for (const key in json) {
+            if (json.hasOwnProperty(key)) {
+                arr.push(TableController.jsonKeyValueToArray(key, json[key]));
+            }
+        }
+        return arr;
+    };
+}
```
IsaackMwenda: nit: add a new line at the end you can also setup vs-code to always do this for you  https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files

public/table.js
```diff
@@ -0,0 +1,106 @@
...
+        let data = codaData.data,
+            lastUpdate = codaData.lastUpdate; 
+
+        // Set last updated timestamp in UI
+        document.getElementById("last-update").innerText = `Last updated: ${lastUpdate}`;
+
+        // Save sorting state
+        let sortInfo = { column: "", order: "" };
+
+        // Invoke `transform` function with column to be sorted on page load
```
IsaackMwenda: This works well on page load, but when you sort by a different column e.g NC% and firebase updates the data  (which happens every 10 sec or so) the sorting goes back to default. It would be great if we can maintain user sorting when data updates.

IsaackMwenda (changes requested):

public/table.js
DanielMwendwa: Thanks, I've enable it

DanielMwendwa:
Made the requested changes above

public/table.js
IsaackMwenda: @DanielMwendwa  This has not been resolved yet.

IsaackMwenda (changes requested):
Requesting changes because on data update user sorting is replaced by default sorting ('Done').

DanielMwendwa:
> Requesting changes because on data update user sorting is replaced by default sorting ('Done').

For me it's just working as intended

DanielMwendwa:
> > Requesting changes because on data update user sorting is replaced by default sorting ('Done').
> 
> For me it's just working as intended
Eg, below have sorted the table by column Dataset
Screen Shot 2020-05-05 at 13 19 01
Screen Shot 2020-05-05 at 13 22 13

public/table.js
DanielMwendwa: Have you tested the branch with the changes below?

IsaackMwenda: Your right I forgot to pull the new changes. This looks okay now thanks.

IsaackMwenda (approved):

public/codingprogress.js
```diff
@@ -0,0 +1,113 @@
+class TableController {
```
as2388:
```suggestion
class CodingProgressTableController {
```

public/codingprogress.js
```diff
@@ -0,0 +1,113 @@
...
+            TableController.column = sortInfo.column;
+
+            d3.select("tbody").selectAll("tr").remove();
+
+            // Table Header
+            d3.select("thead").selectAll("th")
+                .data(TableController.jsonToArray(data[0]))
+                .enter().append("th")
+                .attr("class", "table-heading")
+                .on("click", (d, i, n) => transform(d[0]))
```
as2388:
```suggestion
                .on("click", (d) => transform(d[0]))
```
Then I don't have to think about what irrelevant args mean

public/styles.css
```diff
@@ -206,6 +206,12 @@ table > tbody > tr > td {
     border-radius: 8px;			
     pointer-events: none;			
 }
+tr:hover td { 
+    cursor: pointer;
+}
+th:hover {
+    cursor: grab;
```
as2388: I think this should be pointer not grab, because it's something that's clickable not something that's draggable.

as2388: (But great that the cursor now changes!)

public/index.html
```diff
@@ -53,6 +53,7 @@
     <script src="data.js"></script>
     <script src="UI.js"></script>
     <script src="graph.js"></script>
+    <script src="codingprogress.js"></script>
```
as2388:
```suggestion
    <script src="coding_progress.js"></script>
```
or
```suggestion
    <script src="codingProgress.js"></script>
```

public/codingprogress.js
```diff
@@ -0,0 +1,113 @@
...
+                .text(d => d[0])
+
+            // Table Rows
+            let tr = d3.select("tbody").selectAll("tr")
+                .data(data)
+                .enter().append("tr")
+                .sort((a, b, order = sortInfo.order) => 
+                    a == null || b == null ? 0 : column == "Dataset" ?
```
as2388: This is slightly unsafe because it requires you to call your string column "Dataset" and assumes everything else is a number, which might not be true in the data the caller provides. It might be safer to call sortNumber if both values have type number, otherwise fall back to string.

as2388 (approved):
Very nice :)

Is there any old table-sorting code/library imports that can be removed?

DanielMwendwa:
> Very nice :)
> 
> Is there any old table-sorting code/library imports that can be removed?

Yeah jquery tablesorter. I'll remove it

public/codingprogress.js
DanielMwendwa: Thanks @as2388, I have noted my assumptions on working with numbers were wrong.
please check 8db9ef5 and bfafb59.

PR #212, opened 2020-04-29
Tooltips for 10min view

### LINKED ISSUE #208: Add Tooltips to 10 minute view
The new tooltips look great, but for some reason only show on the 1-day view graphs. I think it would be helpful to show them on the 10-minute view graphs too, particularly with a timestamp showing the time range each bar covers.

The underlying problem I have is that it's very hard to tell when in the day a particular bar happened by eye, which makes it hard to answer questions like "when in the show did data start spiking?", or "when did we last receive a message from this operator?". I will happily take an alternative solution to that problem if needed.

DanielMwendwa:
Closes: #208
Copied code from stacked bar graphs' tooltip implementation and added time of the day to the tooltips in 10 min view as suggested in issue above.

### DIFF
--- a/public/graph.js
+++ b/public/graph.js
@@ -374,16 +374,16 @@ class GraphController {
                 .selectAll("rect")
                 .on("mouseover", (d, i, n) => {
                     // Get key of stacked data from the selection
-                    let operatorReceived = d3.select(n[i].parentNode).datum().key,
+                    let operatorNameWithMessageDirection = d3.select(n[i].parentNode).datum().key,
                         // Get operator name from the key
-                        operatorName = operatorReceived.replace('_received',''),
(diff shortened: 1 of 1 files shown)


public/graph.js
```diff
@@ -17,6 +17,7 @@ class GraphController {
             isYLimitSentManuallySet = false,
             isYLimitFailedManuallySet = false,
             dayDateFormat = d3.timeFormat("%Y-%m-%d"),
+            dayTimeFormat = d3.timeFormat("%H:%M %p"),
```
IsaackMwenda, line 20: please investigate what's causing this.
image

IsaackMwenda (approved):
LGTM!

public/graph.js
DanielMwendwa: I will

PR #213, opened 2020-04-30
Tooltip stuck

DanielMwendwa:


DanielMwendwa:
I had checked out this branch from another branch in review instead of master.

PR #214, opened 2020-04-30
RESOLVE: Tooltip doesn't disappear on mouseout after taking long

### LINKED ISSUE #205: Tooltip doesn't disappear on mouseout after taking long
The tooltip remains stuck on the graph after taking long to hover out of a bar.
tooltip

DanielMwendwa:
On mouse out the tooltip doesn't disappear after browser taking long open. Noted this doesn't appear on safari. Resolved this on other browsers by passing node hovered and its bound data to tip.hide function on mouse out to hide the tooltip on mouse out.

Closes: #205

### DIFF
--- a/public/graph.js
+++ b/public/graph.js
@@ -490,7 +490,7 @@ class GraphController {
                     tip.show(d, n[i]).style("color", operatorColor)
                 })
                 .on("mouseout", (d, i, n) => {
-                    tip.hide()
+                    tip.hide(d, n[i])
                 })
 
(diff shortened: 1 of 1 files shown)


IsaackMwenda (commented):
May I check how you have tested this @DanielMwendwa ?

I left the mouse pointer over a segment of a bar chart, left it for around 20 minutes, then came to check if the tooltip stuck.

IsaackMwenda (approved):

PR #220, opened 2020-05-05
Compute x axis tick values for sent & failed 10 min view graph

DanielMwendwa:
Replicated changes on approved PR #219


IsaackMwenda (approved):
As discussed offline the best approach for such cases is to revert the commits and push to a different branch.

PR #221, opened 2020-05-05
Add grid lines to received msg graph

DanielMwendwa:
Added grid lines to received messages graph

### DIFF
--- a/public/graph.js
+++ b/public/graph.js
@@ -339,6 +339,7 @@ class GraphController {
             d3.selectAll(".redrawElementReceived").remove();
             d3.selectAll("#receivedStack").remove();
             d3.selectAll("#receivedStack10min").remove();
+            d3.selectAll(".receivedGrid").remove();
 
             // Group data filtered by week daily and generate tick values for x axis
             let dataFilteredWeekGroupedDaily  = d3.nest().key(d => d.day)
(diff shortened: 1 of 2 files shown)


public/styles.css
```diff
@@ -206,3 +206,13 @@ table > tbody > tr > td {
     border-radius: 8px;			
     pointer-events: none;			
 }
+.receivedGrid line {
```
IsaackMwenda, line 212: a universal class for all the graphs would be cleaner.

public/styles.css
```diff
@@ -206,3 +206,13 @@ table > tbody > tr > td {
...
+    stroke: lightgrey;
+    stroke-opacity: 0.7;
```
IsaackMwenda: make it more transparent so that the graphs don't look too busy

public/graph.js
```diff
@@ -360,6 +361,24 @@ class GraphController {
             }
             const tickValuesForXAxis = dataFilteredWeekGroupedDaily.map(d => d.datetime);
 
+            // Add the X gridlines
```
IsaackMwenda, line 366: This makes more sense on 10 minutes view, the 1-day gridlines view makes the graph look too busy and congested, after all we have spacing between the bars so its already clear where each day starts.

IsaackMwenda (approved):

public/styles.css
DanielMwendwa: Yeah but it won't work because I am using the class to remove available gridlines when one clicks on y control button `d3.selectAll(".receivedGrid").remove();` without which the gridlines will fill the graph area.  This explains why each graph gridlines' should be independent of others.

public/graph.js
DanielMwendwa: I will reduce the opacity of x gridlines in one day view

public/styles.css
DanielMwendwa: Done

PR #222, opened 2020-05-06
Experiment Clickable stacked bar chart on One day view received graph 

DanielMwendwa:


DanielMwendwa:
Modularise stacked bar chart first

PR #228, opened 2020-05-13
RESOLVE: Message coding table resets on data update

### LINKED ISSUE #226: Message coding table resets on data update
Coding progress table sorting state resets to the default state of "Done" column being sorted in descending order instead of the current sort state defined by the user.

DanielMwendwa:
Closes #226 
1. Added an external function to save sorting information which receives the column to be sorted as its parameter.
2. Added an optional parameter to function used to generate coding progress table 
3. The parameter is used on click events of header/column of the table to pass updated sorting information 
4. The updated sorting info is generated after saving sorting information 

To check how this works checkout branch `cp-table-reset-issue2` and view console logs while clicking on the table header/column


DanielMwendwa:
The code was working right but it could be hard to understand how it works because there was a logical error from sort order description, I was regarding descending data as ascending and vice versa. Resolved: 87f674f

IsaackMwenda (approved):
LGTM thanks.

PR #231, opened 2020-05-13
Add auth css file

DanielMwendwa:
I've separated code that is rarely used to minimize the size of style.css file & also because the are used independently.

### DIFF
--- a/public/auth.css
+++ b/public/auth.css
@@ -1,0 +1,34 @@
+.login-body {
+    margin: 0;
+    padding: 0;
+    background: #ffff;
+    background-size: cover;
+    background-position: center;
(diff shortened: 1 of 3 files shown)


IsaackMwenda:
Hi @DanielMwendwa  there are a few different factors to consider why a single css file is better than multiple:

- Its easier to maintain one file than multiple
- HTTP request times - each CSS file is a separate request, which adds on a little overhead for page load time.
- Browser caching - if the browser has already cached the CSS used on index.html, you don’t want to make it load a “new” CSS file for auth.html.

I would recommend keeping the single css file since its easy to maintain one file and we don't have any overlapping css classes

PR #233, opened 2020-05-14
RESOLVE: Message coding table sorts on row click

### LINKED ISSUE #232: Message coding table sorts on row click
Clicking any row of the message coding table triggers a sort on that column, when I think we just want to sort if the table header is clicked.

This is especially problematic when opening a dataset - clicking a dataset opens the dataset and sorts the table, when it should only open the dataset.

DanielMwendwa:
Closes: #232


as2388 (approved):
Thanks @DanielMwendwa.

I think you should also remove the pointer: cursor from the table rows, now that only the links do anything on click.

PR #234, opened 2020-05-17
system-monitoring-charts 

DanielMwendwa:


DanielMwendwa:
All the commits hereafter will be those of generating the graphs with d3 .

DanielMwendwa:
All the commits hereafter will be those of generating the graphs with d3.

DanielMwendwa:
Generated the first graph for disk usage metrics.

public/UI.js
```diff
@@ -3,6 +3,7 @@ class UIController {
     static getDOMstrings() {
         return {
             projectMenu: ".project-menu",
+            systemsLinkSelector: ".systems-link",
```
IsaackMwenda: move to the last item in the menu.

public/index.html
```diff
@@ -51,7 +52,8 @@
     <script src="auth.js"></script>
     <script src="data.js"></script>
     <script src="UI.js"></script>
-    <script src="graph.js"></script>
+    <script src="traffic_graphs.js"></script>
+    <script src="systems_graphs.js"></script>
```
IsaackMwenda:
```suggestion
    <script src="systems_monitoring_graphs.js"></script>
```

public/systems_graphs.js
```diff
@@ -0,0 +1,209 @@
...
+            // Create keys to stack
+            let diskKeys = [],
+            diskStr = ""
+
+            for (let i = 0; i < diskMetrics.length; i++) {
+                diskStr =  "disk_" + diskMetrics[i];
+                diskKeys.push(diskStr);
+            }
+
+            let colors = ["#228B22", "#fbdb9c"]
```
IsaackMwenda: try #0E86D4 for used and #228B22 for free (with current opacity)

public/systems_graphs.js
```diff
@@ -0,0 +1,209 @@
...
+
+            // Add X axis label
+            svg.append("text")
+                .attr("text-anchor", "end")
+                .attr(
+                    "transform",
+                    "translate(" + Width / 2 + " ," + (Height + Margin.top + 50) + ")"
+                )
+                .style("text-anchor", "middle")
+                .text("Date (dd HH:mm)");
```
IsaackMwenda: nit: 
```suggestion
                .text("Date (dd:h:m)");
```

public/systems_graphs.js
```diff
@@ -0,0 +1,209 @@
...
+    static updateGraphs(data) {
+
+        let dayTimeFormat = d3.timeFormat("%a %d (%H:%M)");
+
+        // Clear previous graphs before redrawing
+        d3.selectAll("svg").remove();
+
+        // Set the dimensions and margins of the graphs
+        const Margin = {top: 40, right: 100, bottom: 105, left: 70},
+            Width = 960 - Margin.left - Margin.right,
```
IsaackMwenda: The graph is drawn over the y-axis line.

public/systems_graphs.js
```diff
@@ -0,0 +1,209 @@
...
+                
+            let yLimit = data[0].disk_total
+            // Add Y axis
+            let y = d3.scaleLinear()
+                .domain([0, yLimit])
+                .range([ Height, 0 ]);
+                svg.append("g")
+                .call(d3.axisLeft(y).ticks(5))
+
+            // Add a clipPath: everything out of this area won't be drawn.
```
IsaackMwenda: Suggestion: it is probably better to only create an MVP with no new features from what we currently have in other graphs till you finish refactoring the code. clipPath here is great but I think it can wait for that. I will leave it to you to decide.

IsaackMwenda: Tooltips would be great to have them now though.

public/index.html
```diff
@@ -51,7 +52,8 @@
     <script src="auth.js"></script>
     <script src="data.js"></script>
     <script src="UI.js"></script>
-    <script src="graph.js"></script>
+    <script src="traffic_graphs.js"></script>
```
IsaackMwenda:
```suggestion
    <script src="traffic_monitoring_graphs.js"></script>
```

IsaackMwenda (changes requested):
Thanks @DanielMwendwa, requesting changes to look at the other graphs code once you add them

public/systems_graphs.js
DanielMwendwa: I'm working on the tooltip, it may take a little bit longer.

DanielMwendwa: If we remove clip path, we should remove also remove zooming effect because the work together.

IsaackMwenda: I was referring to both here. but happy for you to keep it if it does not make your work harder when refactoring.

public/UI.js
IsaackMwenda: I think this is not resolved yet.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,576 @@
...
+            // Create keys to stack
+            let memoryKeys = [],
+            memoryStr = ""
+
+            for (let i = 0; i < memoryMetrics.length; i++) {
+                memoryStr =  "memory_" + memoryMetrics[i];
+                memoryKeys.push(memoryStr);
+            }
+
+            let colors = ["#4B0082", "#fbdb9c"]
```
IsaackMwenda, line 217: maybe use the same color scheme as the disc graph? happy for you to decide.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,576 @@
...
+                .attr("x", Width / 2)
+                .attr("y", 0 - Margin.top / 2)
+                .attr("text-anchor", "middle")
+                .style("font-size", "20px")
+                .style("text-decoration", "bold")
+                .text("Disk Usage");
+
+        }
+
+        function plotMemoryMetrics(data) {
```
IsaackMwenda, line 368: Apart from disc usage; We should just plot used metric only for both memory and CPU utilization. Free disc space varies every time we delete files, but total memory & CPU metrics are constant so we can visually estimate the amount available.

public/UI.js
DanielMwendwa: resolved

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,576 @@
...
+
+            // Add X axis label
+            svg.append("text")
+                .attr("text-anchor", "end")
+                .attr(
+                    "transform",
+                    "translate(" + Width / 2 + " ," + (Height + Margin.top + 50) + ")"
+                )
+                .style("text-anchor", "middle")
+                .text("Date (dd:h:m)");
```
IsaackMwenda:
```suggestion
                .text("Date (dd:hh:m)");
```

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,576 @@
...
```
IsaackMwenda:
```suggestion
                .text("Date (dd:hh:m)");
```

IsaackMwenda (changes requested):

public/systems_graphs.js
DanielMwendwa: I will handle those features in the refactoring, The zooming feature helps to look at the data at specified times of the day.

public/systems_monitoring_graphs.js
DanielMwendwa: I'll just remove them from the stack, therefore after refactoring I will just have one function that can plot all the three graphs using different dataset

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+                    .attr("class", d => "diskArea " + d.key)
+                    .style("fill", d => color(d.key))
+                    .attr("d", area)
+                    .on("mouseover", function() {
+                        tooltip.style("display", null);
+                    })
+                    .on("mouseout", function() {
+                    tooltip.style("display", "none");
+                    })
+                    .on("mousemove", function(d) {
```
IsaackMwenda: I think we should restrict tooltip to used graph only, the free space graph here is more of a visual enhancement so we don't need a tooltip for it. It is also not easy to interpret the tooltip for it because the values pick from o instead of max y value.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+            }
+
+            let colors = ["#0E86D4", "#bdefbd"]
+            // color palette
+            let color = d3.scaleOrdinal().domain(diskMetrics).range(colors);
+
+            let stackDisk = d3.stack().keys(diskKeys),
+                diskDataStacked = stackDisk(data);
+
+            let svg = d3.select(".chart")
```
IsaackMwenda: add comment please

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+            // Add a clipPath: everything out of this area won't be drawn.
+            let clip = svg.append("defs").append("svg:clipPath")
+                .attr("id", "clip")
+                .append("svg:rect")
+                .attr("width", Width )
+                .attr("height", Height )
+                .attr("x", 1)
+                .attr("y", 0);
+
+            // Add brushing
```
IsaackMwenda: zooming does not seem to work for its graph, please investigate.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+                .text("Date (dd:hh:m)");
+
+            // Add Y axis label
+            svg.append("text")
+                .attr("transform", "rotate(-90)")
+                .attr("y", 0 - Margin.left)
+                .attr("x", 0 - Height / 2)
+                .attr("dy", "1em")
+                .style("text-anchor", "middle")
+                .text("GB")
```
IsaackMwenda:
```suggestion
                .text("Total Memory Used (GB)")
```

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+
+            // Disk usage graph title
+            svg.append("text")
+                .attr("x", Width / 2)
+                .attr("y", 0 - Margin.top / 2)
+                .attr("text-anchor", "middle")
+                .style("font-size", "20px")
+                .style("text-decoration", "bold")
+                .text("Disk Usage");
+
```
IsaackMwenda:
```suggestion
```

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+                .text("Date (dd:hh:m)");
+
+            // Add Y axis label
+            svg.append("text")
+                .attr("transform", "rotate(-90)")
+                .attr("y", 0 - Margin.left)
+                .attr("x", 0 - Height / 2)
+                .attr("dy", "1em")
+                .style("text-anchor", "middle")
+                .text("GB")
```
IsaackMwenda:
```suggestion
                .text("Disc Usage (GB)")
```

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+                .text("Date (dd:hh:m)");
+
+            // Add Y axis label
+            svg.append("text")
+                .attr("transform", "rotate(-90)")
+                .attr("y", 0 - Margin.left)
+                .attr("x", 0 - Height / 2)
+                .attr("dy", "1em")
+                .style("text-anchor", "middle")
+                .text("%")
```
IsaackMwenda:
```suggestion
                .text("CPU Utilization (%)")
```

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+                        .tickFormat("")
+                    ) 
+            }
+
+            // Memory usage graph title
+            svg.append("text")
+                .attr("x", Width / 2)
+                .attr("y", 0 - Margin.top / 2)
+                .attr("text-anchor", "middle")
+                .style("font-size", "20px")
```
IsaackMwenda, line 685: Try and move most of the duplicated styles to styles.css file

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+                .on("mouseout", function() {
+                    tooltip.style("display", "none");
+                })
+                .on("mousemove", function(d) {
+                    var xPosition = d3.mouse(this)[0] - 15;//x position of tooltip
+                    var yPosition = d3.mouse(this)[1] - 25;//y position of tooltip
+                    tooltip.attr("transform", "translate(" + xPosition + "," + yPosition + ")");//placing the tooltip
+                    var x0 = x.invert(d3.mouse(this)[0]);//this will give the x for the mouse position on x
+                    var y0 = y.invert(d3.mouse(this)[1]);//this will give the y for the mouse position on y
+                    tooltip.select("text").text(d3.timeFormat('%Y-%m-%d')(x0)+ " " +Math.round(y0));//show the text after formatting the date
```
IsaackMwenda: Date {
```suggestion
                    tooltip.select("text").text(d3.timeFormat('%Y-%m-%d')(x0)+ " " Used: +Math.round(y0));//show the text after formatting the date
```

IsaackMwenda: same for other graphs

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+
+            // Add Y axis
+            let y = d3.scaleLinear()
+                .domain([0, 100])
+                .range([ Height, 0 ]);
+            svg.append("g").call(d3.axisLeft(y).ticks(5))
+
+            // Add the Y gridlines
+            svg.append("g")			
+                .attr("class", "diskGrid")
```
IsaackMwenda: rename this class since its being used in all the charts, not disk only.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+            areaChart
+                .append("g")
+                .attr("class", "brush")
+                .call(brush);
+
+            // Add the area
+            areaChart.append("path")
+                .datum(data)
+                .attr("class", "memoryArea")  // I add the class memoryArea to be able to modify it later on.
+                .attr("fill", "yellow")
```
IsaackMwenda: I prefer the previous blue color

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,636 @@
...
+                .attr("fill-opacity", .6)
+                .attr("stroke", "black")
```
IsaackMwenda: this collides with the gridline colors, please make it uniform with the area color.

IsaackMwenda (changes requested):
Thanks for the improvements @DanielMwendwa. Mostly smaller tweaks I'd like to check once addressed.

public/systems_monitoring_graphs.js
DanielMwendwa: The zooming happens outside the area charts, they can't coexist, try zooming above the area chart.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,641 @@
...
+                .text("Date (dd:hh:m)");
+
+            // Add Y axis label
+            svg.append("text")
+                .attr("transform", "rotate(-90)")
+                .attr("y", 0 - Margin.left)
+                .attr("x", 0 - Height / 2)
+                .attr("dy", "1em")
+                .style("text-anchor", "middle")
+                .text("Total Memory Used (%)");
```
IsaackMwenda:
```suggestion
                .text("Memory Utilization(GB)");
```

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,641 @@
...
+            areaChart
+                .append("g")
+                .attr("class", "brush")
+                .call(brush);
+
+            // Add the area
+            areaChart.append("path")
+                .datum(data)
+                .attr("class", "memoryArea")  // I add the class memoryArea to be able to modify it later on.
+                .attr("fill", "#000080")
```
IsaackMwenda: a different shade for memory and cpu would be great.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,641 @@
...
+                .attr("height", Height + Margin.top + Margin.bottom)
+                .append("g")
+                .attr("transform",
+                    "translate(" + Margin.left + "," + Margin.top + ")");
+
+            // Add X axis
+            let x = d3.scaleTime().domain(d3.extent(data, d => new Date(d.datetime))).range([1, Width]);
+            let xAxis = svg.append("g")
+                .attr("transform", "translate(0," + Height + ")")
+                .call(d3.axisBottom(x).tickFormat(dayTimeFormat).ticks(13))
```
IsaackMwenda: This still doesn't seem to be okay. The graph should be per day as the data streams in e.g 10 mins timescale of traffic graphs but now in realtime.

IsaackMwenda (changes requested):

DanielMwendwa:
Resolved the issues raised above

DanielMwendwa:
Resolved the issues raised above

public/systems_monitoring_graphs.js
DanielMwendwa: Updated thanks

public/systems_monitoring_graphs.js
DanielMwendwa: Used brown for memory utilisation graph & left CPU utilisation graph as is.

public/systems_monitoring_graphs.js
DanielMwendwa: Updated x axis time scale to be in per day

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,644 @@
...
+                    .on("mouseout", function() {
+                    tooltip.style("display", "none");
+                    })
+                    .on("mousemove", function(d) {
+                        var xPosition = d3.mouse(this)[0] - 15;//x position of tooltip
+                        var yPosition = d3.mouse(this)[1] - 25;//y position of tooltip
+                        tooltip.attr("transform", "translate(" + xPosition + "," + yPosition + ")");//placing the tooltip
+                        var x0 = x.invert(d3.mouse(this)[0]);//this will give the x for the mouse position on x
+                        var y0 = y.invert(d3.mouse(this)[1]);//this will give the y for the mouse position on y
+                        tooltip.select("text").text(`${d3.timeFormat('%Y-%m-%d')(x0)} Used: ${decimalFormatter(y0).replace('G', 'GB')}`);//show the text after formatting the date
```
IsaackMwenda: It would be really useful if we include the hour and minute too.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,644 @@
...
+                    .attr("transform", "translate(0," + Height + ")")
+                    .call(d3.axisBottom(x)
+                        .tickSize(-Height)
+                        .tickFormat("")
+                    )
+            }
+
+            // What to do when one group is hovered
+            let highlight = function(d){
+                console.log(d)
```
IsaackMwenda: remove log statements

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,644 @@
...
+        const Margin = {top: 40, right: 100, bottom: 105, left: 70},
+            Width = 960 - Margin.left - Margin.right,
+            Height = 500 - Margin.top - Margin.bottom;
+
+        plotDiskMetrics(data);
+        plotMemoryMetrics(data);
+        plotCPUMetrics(data);
+
+        function plotDiskMetrics(data) {
+            let diskMetrics = ["used", "free"]
```
IsaackMwenda: Hi @DanielMwendwa Sorry to go round on this again - it's quite hard thinking through the right design and metrics that we actually need to monitor Miranda. After using this for a week now I think we don't need the free metric on disk space too. Used is enough and also gives us the chance to use the brushing feature effectively/consistently as it is on the other graphs.

public/UI.js
```diff
@@ -184,4 +185,22 @@ class UIController {
             .querySelector(DOMstrings.codingProgressContainer)
             .insertAdjacentHTML("beforeend", newHtml);
     }
+
+    static addSystemsGraphs() {
+        let DOMstrings = UIController.getDOMstrings(),
+            html = `<div class="container"> 
+            <section>
+                <div class="card shadow chart my-1"></div>
```
IsaackMwenda:
```suggestion
                <div class="card shadow disc-usage-chart my-1"></div>
```

public/UI.js
```diff
@@ -184,4 +185,22 @@ class UIController {
...
+            </section> 
+            <section>
+                <div class="card shadow chart2 my-1"></div>
```
IsaackMwenda:
```suggestion
                <div class="card shadow memory-utilization-chart my-1"></div>
```

public/UI.js
```diff
@@ -184,4 +185,22 @@ class UIController {
...
+            </section> 
+            <section>
+                <div class="card shadow chart3 my-1"></div>
```
IsaackMwenda:
```suggestion
                <div class="card shadow cpu-utilization-chart my-1"></div>
```

IsaackMwenda (approved):
Thanks for the hard work @DanielMwendwa . These charts will be really useful. Please request a review from @as2388 once you have addressed the minor comments

public/systems_monitoring_graphs.js
DanielMwendwa: Updated thanks.

public/systems_monitoring_graphs.js
DanielMwendwa: removed log statement, thanks

public/UI.js
DanielMwendwa: updated, thanks

public/UI.js
DanielMwendwa: updated, thanks

public/UI.js
DanielMwendwa: updated, thanks

public/systems_monitoring_graphs.js
DanielMwendwa: I understand, updated the stacked area chart to area chart.

DanielMwendwa:
Thanks @IsaackMwenda.

DanielMwendwa:
@IsaackMwenda @as2388, As you'll noticed there is a lot of repetition in the code, now we're plotting three area charts. As discussed earlier on modularizing the code, this will change soon, I've created a class that generates area charts, and will be used instead of the repetition. As from this review lot has changed and I'm working on handling certain features in the area chart class, for now it only generates the area chart only, This is what I'm working on creating [reusable charts](https://github.com/AfricasVoices/Operations-Dashboard/tree/modularize_graph_code/public/avf_graphs).

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,578 @@
...
+                .attr("stroke-width", 0.2)
+                .attr("d", area)
+                .on("mouseover", function() {
+                    tooltip.style("display", null);
+                })
+                .on("mouseout", function() {
+                    tooltip.style("display", "none");
+                })
+                .on("mousemove", function(d) {
+                    var xPosition = d3.mouse(this)[0] - 15;//x position of tooltip
```
as2388:
```suggestion
                    var xPosition = d3.mouse(this)[0] - 15; // x position of tooltip
```
Code is slightly cleaner/easier to read if you put spaces around all of your '//'s. You may be able to configure your editor to warn about this automatically.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,578 @@
...
+                    var yPosition = d3.mouse(this)[1] - 25;//y position of tooltip
+                    tooltip.attr("transform", "translate(" + xPosition + "," + yPosition + ")");//placing the tooltip
+                    var x0 = x.invert(d3.mouse(this)[0]);//this will give the x for the mouse position on x
+                    var y0 = y.invert(d3.mouse(this)[1]);//this will give the y for the mouse position on y
```
as2388: This interpolates rather than using the nearest known value, which I think is a bit misleading because it suggests we're making more frequent readings than we actually are. I think better to get the date/value of the nearest known point.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,578 @@
...
+                    tooltip.select("text").text(`${d3.timeFormat('%Y-%m-%d (%H:%M)')(x0)} Used: ${decimalFormatter(y0).replace('G', 'GB')}`);//show the text after formatting the date
```
as2388: Can you round to the nearest 1GB (or even 0.1) instead of the nearest 10? 10 is quite a lot so it would be good to have a slightly more precise report here

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,578 @@
...
+
+            tooltip.append("text")
+                .attr("x", 95)
+                .attr("dy", "1.2em")
+                .style("text-anchor", "middle")
+                .attr("font-size", "12px")
+                .attr("font-weight", "bold");
+
+            let idleTimeout
+            function idled() { idleTimeout = null; }
```
as2388: Could you add a comment explaining what these are for? They're a little bit too cryptic until you read L164.

as2388: It may also help to say what is being idled here, or call it e.g. transitionDelay.
Idle implies the user is idle i.e. not interacting, but I think all that's going on here is you're waiting before running an animation.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,578 @@
...
+            let clip = svg.append("defs").append("svg:clipPath")
+                .attr("id", "clip")
+                .append("svg:rect")
+                .attr("width", Width )
+                .attr("height", Height )
+                .attr("x", 1)
+                .attr("y", 0);
+
+            // Add brushing
+            let brush = d3.brushX()                 // Add the brush feature using the d3.brush function
```
as2388: Can you move this to be next to where you use it (L295)? It's a bit easier to read if the declaration and use are nearby

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,578 @@
...
+                .attr("clip-path", "url(#clip)")
+
+            // Create an area generator
+            let area = d3.area().x(d => x(d.datetime)).y0(y(0)).y1(d => y(d.memory_usage.used))
+
+            // Add the brushing
+            areaChart
+                .append("g")
+                .attr("class", "brush")
+                .call(brush);
```
as2388: The brush only works when dragging over the white parts of the graph - it should support the filled parts too (i.e. the bits below the line)

as2388: The brush only works when dragging over the white parts of the graph - it should support the filled parts too (i.e. the bits below the line)

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,578 @@
...
+            let brush = d3.brushX()                 // Add the brush feature using the d3.brush function
+                .extent( [ [0,0], [Width, Height] ] ) // initialise the brush area: start at 0,0 and finishes at width,height: it means I select the whole graph area
+                .on("end", updateChart) // Each time the brush selection changes, trigger the 'updateChart' function
+
+            // Create the scatter variable: where both the circles and the brush take place
+            let areaChart = svg.append('g')
+                .attr("clip-path", "url(#clip)")
+
+            // Create an area generator
+            let area = d3.area().x(d => x(d.datetime)).y0(y(0)).y1(d => y(d.memory_usage.used))
```
as2388, line 500: I think we should draw attention to places where we didn't receive data e.g. because the service that was generating it was down, rather than just interpolate it through it. For example, if the tool goes down you could fill the background in grey or show 0s if that was appropriate. At the moment, if the tool goes down you see periods of straight lines like these:
image
it took me a while to figure out why our pipelines were behaving so weirdly before realising it wasn't actually the pipelines that were the problem

public/traffic_monitoring_graphs.js
```diff
@@ -0,0 +1,1294 @@
+// GRAPH CONTROLLER
+class GraphController {
```
as2388, line 2: Am I right this has just moved from somewhere else? If so, does it need removing from that old place?

public/systems_monitoring_graphs.js
as2388: Otherwise this feature is really great! Have you considered adding it to the traffic graphs too? It would be really helpful to be able to zoom in on those, and I think this would work as a much nicer alternative to the existing time-frame drop-down.

public/index.html
```diff
@@ -51,7 +52,8 @@
     <script src="auth.js"></script>
     <script src="data.js"></script>
     <script src="UI.js"></script>
-    <script src="graph.js"></script>
+    <script src="traffic_monitoring_graphs.js"></script>
+    <script src="systems_monitoring_graphs.js"></script>
```
as2388, line 56: The line here is twice as thick as all the others:
image
I think we had this problem before, so hopefully it's straightforward to apply the same solution again here

public/UI.js
```diff
@@ -184,4 +185,22 @@ class UIController {
             .querySelector(DOMstrings.codingProgressContainer)
             .insertAdjacentHTML("beforeend", newHtml);
     }
+
+    static addSystemsGraphs() {
+        let DOMstrings = UIController.getDOMstrings(),
+            html = `<div class="container">
```
as2388, line 191: Can you add last updated timestamp (and red warning) just like we have for the traffic graphs?

as2388 (changes requested):
Thanks @DanielMwendwa these look great, and the code was easy to follow too :)

I've added a few comments - some make sense to be addressed now but there are a few which you might find easier to convert into issues and address later. Because these charts are really only for you, me and Isaack, we could probably deploy slightly earlier than we would usually because the risk of us misinterpreting the charts is much lower, which would really help with our day-to-day monitoring work.

public/systems_monitoring_graphs.js
DanielMwendwa: Moved all the brush variable declaration next to where its used, thanks.

public/index.html
DanielMwendwa: Updated the right border position, thanks

public/systems_monitoring_graphs.js
DanielMwendwa: Updated thanks. In the code, idleTimeout will hold time set before subsequent brushing trigger `updateChart` function.
i.e.,
```
var returnObj = setTimeout(
    function(){console.log('ping');}, 
    1000
);
console.log(returnObj);

```
When tested with node.js, it returns object with the following property
`{ _idleTimeout: 1000, .... }`
That's why have used idleTimeout cause it contains the value of the second arg to set timeout - The number of milliseconds to wait before executing the code

public/UI.js
DanielMwendwa: I've added

public/systems_monitoring_graphs.js
DanielMwendwa: Updated thanks.

public/systems_monitoring_graphs.js
DanielMwendwa: Updated tooltip message.

public/systems_monitoring_graphs.js
DanielMwendwa: I also wanted this but the tooltip & brushing cannot coexist, both of them contain an overlay where the tooltip/brushing happens, when they overlap only the one on top works.

public/systems_monitoring_graphs.js
DanielMwendwa: Updated thanks.

public/systems_monitoring_graphs.js
DanielMwendwa: Yes, I am planning implementing it.

public/systems_monitoring_graphs.js
DanielMwendwa: Let me work on this later

public/traffic_monitoring_graphs.js
DanielMwendwa: yeah, I'll remove it, if you delete graph.js file, it leads to conflict with master which can't be resolved.

DanielMwendwa:
@as2388 I have made substantial changes on the tooltip, I would like you to take a look, as I work on [this](https://github.com/AfricasVoices/Operations-Dashboard/pull/234#discussion_r431120733)

public/systems_monitoring_graphs.js
DanielMwendwa: May be they can, let me work on it one more time...

DanielMwendwa: I was able to solve this. The key to making them coexist is to add the brush (and allow it to create its rect) and then use that rect to add the tooltip events. This way you only end up with one point-events rect

as2388:
> @as2388 I have made substantial changes on the tooltip, I would like you to take a look, as I work on [this](https://github.com/AfricasVoices/Operations-Dashboard/pull/234#discussion_r431120733)

Commenting only on the tooltip: I think this is better than before as there's less opportunity for misreading. You might like to make the circle a lot smaller and apply the same styles you were using for the old tool tip text and container. There also seems to be a performance issue on the disk usage graph for me, which is surprising because the other graphs are super responsive.

DanielMwendwa:
@as2388 I've made the requested changes, please take a look

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,1087 @@
...
+                .call(d3.axisLeft(y)
+                    .tickSize(-Width)
+                    .tickFormat("")
+                )
+
+            // Get time when the system stopped & when it was restarted
+            let stop = [], start = [], currentTime = new Date();
+            data.forEach((d, i, n)=> {
+                if (i == (n.length-1)) {
+                    let difference_ms = (currentTime.getTime() - d.datetime.getTime()) / 60000,
```
as2388: Is this right? If you divide by 60k I think you're computing the difference_minutes, not difference_ms.

That would explain why you missed this outage:
image

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,1087 @@
...
+                    difference_minutes = Math.floor(difference_ms % 60);
+                    if (difference_minutes > 10) {
```
as2388: Assuming the minutes calculation is now correct - I think the watcher is now sleeping for 10 minutes between runs (check with the KK team if unsure) - so we should wait longer before reporting an outage e.g. 30 minutes.

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,1087 @@
...
+            // Add a clipPath: everything out of this area won't be drawn.
+            svg.append("defs").append("svg:clipPath")
+                .attr("id", "clip-lines")
+                .append("svg:rect")
+                .attr("width", Width )
+                .attr("height", Height )
+                .attr("x", 1)
+                .attr("y", 0);
+
+            let hitchLines = svg.append('g')
```
as2388: Slight preference for outageLines, because I think that's a little more descriptive of the type of hitch we've encountered

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,1087 @@
...
+                .attr("y", 0);
+
+            // Create the scatter variable: where both the circles and the brush take place
+            let areaChart = svg.append('g')
+                .attr("clip-path", "url(#clip)")
+            
+            // Area generator
+            let area = d3.area().x(d => x(d.datetime)).y0(y(0))
+                .y1(d => y(d.disk_usage.used))
+                // .curve(d3.curveCardinal); // this smooths out the curves of the line
```
as2388: I think this can be removed

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,1087 @@
...
+                    let i = bisectDate(data, x0, 1);
+                    let d0 = data[i - 1];
+                    let d1 = data[i];
+                    let d = x0 - d0.datetime > d1.datetime - x0 ? d1 : d0;
+    
+                    // Place the focus objects on the same path as the line
+                    focus.attr("transform", `translate(${x(d.datetime)}, ${y(d.disk_usage.used)})`);
+    
+                    // Position the text
+                    focus.select("text")
```
as2388, line 223: Can you update this to show the nearest data point (i.e. data for the point highlighted by the circle) instead of the interpolated timestamps and values?

public/systems_monitoring_graphs.js
```diff
@@ -0,0 +1,1087 @@
...
+                svg.append("g")			
+                    .attr("class", "diskXGrid")
+                    .attr("transform", "translate(0," + Height + ")")
+                    .call(d3.axisBottom(x)
+                        .tickSize(-Height)
+                        .tickFormat("")
+                    )
+
+                // Add the hitch gridlines
+                hitchLines.append("g")
```
as2388: Low priority, but the hitch lines aren't transitioning with the rest of the graph when brush-zooming

as2388 (approved):
Thanks @DanielMwendwa, these look great!

You changed a bit more than I was expecting since my last review, so could you ask @IsaackMwenda to take one final quick look when you've addressed my comments?

public/systems_monitoring_graphs.js
DanielMwendwa: Updated thanks.

public/systems_monitoring_graphs.js
DanielMwendwa: Updated, thanks.

public/systems_monitoring_graphs.js
DanielMwendwa: Updated thanks.

public/systems_monitoring_graphs.js
DanielMwendwa: Updated thanks

public/systems_monitoring_graphs.js
DanielMwendwa: Thanks, updated the variable.

DanielMwendwa:
Thanks @as2388.

DanielMwendwa:
Hi @IsaackMwenda please take a look before I merge this branch.

public/styles.css
```diff
@@ -229,3 +230,47 @@ th:hover {
...
+}
+.stopDiskGrid path,
+.stopMemoryGrid path,
+.stopCpuGrid path,
+.startDiskGrid path,
+.startMemoryGrid path,
+.startCpuGrid path {
+    stroke-opacity: 0;
+}
+
```
IsaackMwenda: remove trailing white spaces

IsaackMwenda (approved):
LGTM, thanks for the improvements @DanielMwendwa  @as2388.

public/styles.css
DanielMwendwa: Updated thanks.

PR #237, opened 2020-05-26
Javascript Modules

DanielMwendwa:
Why use modules? Maintainability,  Namespacing (modules allow us to avoid namespace pollution by creating a private space for our variables), Reusability. more info on [modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules)


IsaackMwenda (changes requested):
Hi Daniel please resolve conflicts

DanielMwendwa:
conflicts resolved

public/app.js
```diff
@@ -1,10 +1,16 @@
...
+import { DataController } from "./data.js";
+
 // GLOBAL APP CONTROLLER
 class Controller {
     static setupEventListeners() {
         Controller.DOMstrings = UIController.getDOMstrings();
         document
             .querySelector(Controller.DOMstrings.logoutBtn)
-            .addEventListener("click", AuthController.logout);
+            .addEventListener("click", () => {import("./auth.js").then(module => {
```
IsaackMwenda: Why are dynamically importing this? It is used twice so cleaner if imported statically.

public/graph.js
```diff
@@ -0,0 +1,1294 @@
+// GRAPH CONTROLLER
+export class GraphController {
```
IsaackMwenda: please delete before merging.

IsaackMwenda (approved):

public/app.js
DanielMwendwa: Updated thanks.

public/graph.js
DanielMwendwa: Deleted  graph.js file

PR #238, opened 2020-05-26
Modularize sys-metrics graph code

### LINKED ISSUE #130: Break down graph update function into different modules

DanielMwendwa:
Please review PR #237 first.
Fix #130 
[more info...](https://github.com/AfricasVoices/Operations-Dashboard/pull/234#issuecomment-634614806)


IsaackMwenda:
Please merge master into this

DanielMwendwa:
> Please merge master into this

Updated thanks.

public/avf_graphs/area.chart.js
```diff
@@ -0,0 +1,427 @@
+import { GraphLayout } from "./graph.layout.js";
```
IsaackMwenda: The number of js files is increasing and having non-descriptive folder names for them will make it hard to maintain. Please; 
- [x] store/move all js files in one `js` folder.
- [x] Rename all js files with respect to the class name e.g UI.js -> UIController.js so that we don't have to remember 2 names.
- [x] Move css file to CSS folder.

IsaackMwenda (changes requested):

public/avf_graphs/area.chart.js
DanielMwendwa: Updated thanks.

DanielMwendwa:
Updated the folders structure

IsaackMwenda (changes requested):
We have duplicate files in different folders under the public folder. Please delete the duplicates. Also, libs and controllers are subdirectories of js folder, not standalone directories.

DanielMwendwa:
> We have duplicate files in different folders under the public folder. Please delete the duplicates. Also, libs and controllers are subdirectories of js folder, not standalone directories.

The duplicate files don't appear on my local repo, I had moved the files but I've been able to delete them here and I have updated the repo, thanks.

IsaackMwenda:
> > We have duplicate files in different folders under the public folder. Please delete the duplicates. Also, libs and controllers are subdirectories of js folder, not standalone directories.
> 
> The duplicate files don't appear on my local repo, I had moved the files but I've been able to delete them here and I have updated the repo, thanks.

It happens. Once you delete locally, you have to push the delete commits so that the change is updated here.

public/js/controllers/traffic_graphs.js
```diff
@@ -1,18 +1,18 @@
 // GRAPH CONTROLLER
-export class GraphController {
+export class TrafficGraphsController {
```
IsaackMwenda: rename the file according to the class name in this case traffic_graphs_controller.js so that we don't have to remember two names.

IsaackMwenda (approved):
This lgtm @DanielMwendwa. with minor comments to rename js files with respect to the class names.
Also, investigate the following error in the debug console before merging
image

public/js/controllers/traffic_graphs.js
DanielMwendwa: Updated thanks

DanielMwendwa:
> This lgtm @DanielMwendwa. with minor comments to rename js files with respect to the class names.
> Also, investigate the following error in the debug console before merging
> image

The error doesn't occur that frequently, I haven't seen it recently, it's caused when one of the values used to compute the height of the bar charts is undefined for example see the statements below; 
`.attr("height", d => y_total_sent_sms_range(d[0]) - y_total_sent_sms_range(d[1]))`
`.attr("height", d => undefined - y_total_sent_sms_range(d[1]))` : This will raise the error above
The problem is why the values become undefined, maybe its internet connection, @IsaackMwenda can you check if the error still appears. For me I can't see the error.
I've also noticed, this error doesn't affect the plotting of the bar charts most of the time.

public/js/controllers/UI.controller.js
```diff
@@ -34,7 +34,7 @@ export class UIController {
     static addCodingProgressSection() {
         let DOMstrings = UIController.getDOMstrings(),
             script = document.createElement('script');
-        script.setAttribute('src','scroll.js');
+        script.setAttribute('src','js/libs/scroll.js');
```
as2388:
```suggestion
        script.setAttribute('src', 'js/libs/scroll.js');
```

public/js/controllers/UI.controller.js
```diff
@@ -109,7 +109,7 @@ export class UIController {
     static addGraphs(title) {
         let DOMstrings = UIController.getDOMstrings(),
             script = document.createElement('script');
-        script.setAttribute('src','scroll.js');
+        script.setAttribute('src','js/libs/scroll.js');
```
as2388:
```suggestion
        script.setAttribute('src', 'js/libs/scroll.js');
```
etc.

public/js/controllers/app.controller.js
```diff
@@ -1,6 +1,6 @@
-import { UIController } from "./UI.js";
-import { DataController } from "./data.js";
-import { AuthController } from "./auth.js";
+import { UIController } from "./UI.controller.js";
```
as2388:
```suggestion
import { UIController } from "./ui_controller";
```
(The '.' is a bit confusing here because . usually separates the name from the type in file names, so better to use '_')

public/js/controllers/app.controller.js
```diff
@@ -23,11 +23,11 @@ class Controller {
     static resetUI() {
         document.querySelector(Controller.DOMstrings.codingProgressContainer).innerHTML = "";
         document.querySelector(Controller.DOMstrings.graphContainer).innerHTML = "";
-        import("./traffic_monitoring_graphs.js").then(module => {
-            module.GraphController.clearTimers();
+        import("./traffic.graphs.controller.js").then(module => {
```
as2388: Same here and everywhere else like this, prefer traffic_graphs_controller to traffic.graphs.controller

IsaackMwenda:
> > This lgtm @DanielMwendwa. with minor comments to rename js files with respect to the class names.
> > Also, investigate the following error in the debug console before merging
> > image
> 
> The error doesn't occur that frequently, I haven't seen it recently, it's caused when one of the values used to compute the height of the bar charts is undefined for example see the statements below;
> `.attr("height", d => y_total_sent_sms_range(d[0]) - y_total_sent_sms_range(d[1]))`
> `.attr("height", d => undefined - y_total_sent_sms_range(d[1]))` : This will raise the error above
> The problem is why the values become undefined, maybe its internet connection, @IsaackMwenda can you check if the error still appears. For me I can't see the error.
> I've also noticed, this error doesn't affect the plotting of the bar charts most of the time.

yup, it is probably an intermittent internet connection. I can't reproduce it now.

public/js/controllers/systems.graphs.controller.js
```diff
@@ -0,0 +1,107 @@
+import { AreaChart } from "../libs/area.chart.js";
+
+export class SystemsGraphsController {
+    static updateGraphs(data) {
+        // Clear previous graphs before redrawing
+        d3.selectAll("svg").remove();
+
+/*                         DISK UTILIZATION GRAPH
+/*      ==================================================================
```
as2388: I don't think this huge block comment is necessary now that's there not much code below, and that the variables have sensible names. I think you can get away with // Draw disk utilization graph or no comment at all. If you think you need a large block comment like this, it's a good sign that the code below should really be in its own function

public/js/controllers/systems.graphs.controller.js
```diff
@@ -0,0 +1,107 @@
...
+*/      let diskUsageChartData = JSON.parse(JSON.stringify(data));
+        diskUsageChartData.forEach(function(d) {
+            d.datetime = new Date(d.datetime);
+            d.value = +d.disk_usage.used;
+        })
+
+        const area3 = new AreaChart(
```
as2388:
```suggestion
        const diskUsageChart = new AreaChart(
```

public/js/libs/area.chart.js
```diff
@@ -0,0 +1,427 @@
...
+        let d1 = this.data[i];
+        let d = x0 - d0.datetime > d1.datetime - x0 ? d1 : d0;
+
+        // Place the focus objects on the same path as the line
+        this.focus.attr("transform", `translate(${this.xScale(d.datetime)}, ${this.yScale(d.value)})`); 
+
+        let tooltipText = `${d3.timeFormat("%Y-%m-%d (%H:%M)")(d.datetime)} 
+        Value: ${d3.formatPrefix(".2", d.value)(d.value)}`;
+
+        if (this.feature == "system-metrics") {
```
as2388: It would be cleaner if you could pass these pieces of configuration in directly rather than switching on the feature/id. That's so that if we need another area chart in future, we don't need to edit various pieces of hard-coded options in the area chart implementation, but instead just pass in the new configuration.

as2388 (approved):

public/js/controllers/UI.controller.js
DanielMwendwa: Updated thanks.

DanielMwendwa: Updated thanks

public/js/controllers/app.controller.js
DanielMwendwa: Updated thanks

public/js/controllers/systems.graphs.controller.js
DanielMwendwa: Updated comments i.e., // Draw disk utilization graph. thanks

public/js/libs/area.chart.js
DanielMwendwa: I find it hard to pass the configuration indirectly. I had this in mind that If we needed another area chart one would not pass the feature setting of which you get an area chart without specific features that system-metrics charts have e.g outage lines.

DanielMwendwa: Updated thanks.

PR #241, opened 2020-06-05
RESOLVE: Message coding table header definition accordion issue

### LINKED ISSUE #225: Message coding table header definition accordion
When you click the accordion it scrolls up then down hiding the content from the view. Ideally, it should scroll up to allow the user to view the content and scroll down on second-click

DanielMwendwa:
Closes: #225


IsaackMwenda (approved):
Thanks

PR #242, opened 2020-06-16
filter sys metrics data by a month

DanielMwendwa:


IsaackMwenda (approved):
That was quick thanks

PR #246, opened 2020-06-24
update mno color scheme to use color for unspecified operator in MNO colors

DanielMwendwa:
…colors

### DIFF
--- a/public/js/controllers/traffic_graphs_controller.js
+++ b/public/js/controllers/traffic_graphs_controller.js
@@ -210,7 +210,8 @@ export class TrafficGraphsController {
         let mno_color_scheme = [],
             operators_with_color_identity = Object.keys(MNOColors);
 
-        let operators_without_color_identity = operators.filter(x => !operators_with_color_identity.includes(x))[0];
+        let operators_without_color_identity = operators.filter(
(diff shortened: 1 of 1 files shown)


public/js/controllers/traffic_graphs_controller.js
```diff
@@ -210,6 +210,10 @@ export class TrafficGraphsController {
         let mno_color_scheme = [],
             operators_with_color_identity = Object.keys(MNOColors);
 
+        let difference = operators.filter(x => !operators_with_color_identity.includes(x))[0];
```
IsaackMwenda: what do you mean by difference?

DanielMwendwa: Operators in array 'operators' that are not in array 'operators_with_color_identity'

IsaackMwenda: Thanks, operators_without_color_identity would be more appropriate.

IsaackMwenda (approved):

public/js/controllers/traffic_graphs_controller.js
DanielMwendwa: Updated thanks

PR #250, opened 2020-06-28
Fix time frame issue  

DanielMwendwa:
The button used to update the time frame on traffic monitoring graphs is not working. I had missed to add an argument that is passed to a function which is executed when the button is clicked.


IsaackMwenda (approved):
Thanks @DanielMwendwa

PR #252, opened 2020-06-29
Update Insertion and removal method of DOM elements

DanielMwendwa:
Updated insertion and removal method of DOM elements to enable efficiency. I.e instead of replacing the inner html of the parent element with an empty string when navigating to another page, I have used an efficient way whereby  I remove the child elements of the parent element using a while loop. I've also come up with some functions to prevent repetition of code.  
This PR also solves the issue where the traffic monitoring graphs page replicates during firestore edits.
duplicate

### DIFF
--- a/public/js/controllers/ui_controller.js
+++ b/public/js/controllers/ui_controller.js
@@ -4,14 +4,12 @@ export class UIController {
         return {
             projectMenu: ".project-menu",
             codingProgressLinkSelector: ".coding-progress-link",
-            codingProgressContainer: ".coding-progress-container",
             trafficsLinkSelector: ".traffics-link",
-            graphContainer: ".graph-container",
+            systemsLinkSelector: ".systems-link",
             logoutBtn: ".logout-btn",
-            dropdownItem: ".dropdown-item",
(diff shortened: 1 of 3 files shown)


IsaackMwenda:
> _No description provided._

Description please

DanielMwendwa:
> > _No description provided._
> 
> Description please

Updated thanks.

public/js/controllers/ui_controller.js
```diff
@@ -31,11 +29,23 @@ export class UIController {
...
+    static resetUI() {
+        const DOMstrings = UIController.getDOMstrings();
+        UIController.statusBody = document.querySelector(DOMstrings.mainContainer);
+        while (UIController.statusBody.firstChild) {
+            UIController.statusBody.removeChild(UIController.statusBody.firstChild);
+        }
+    }
+
+    static getScrollJsScript() {
+        let script = document.createElement('script');
```
IsaackMwenda:
```suggestion
        let ScrollScript = document.createElement('script');
```

IsaackMwenda (approved):

PR #254, opened 2020-06-30
Format code

DanielMwendwa:
Done formatting of code to enable consistency and enhance its readability.

### DIFF
--- a/public/js/libs/firebase.js
+++ b/public/js/libs/firebase.js
@@ -5,6 +5,6 @@ let config = {
     projectId: "avf-dashboards",
     storageBucket: "avf-dashboards.appspot.com",
     messagingSenderId: "1063777163571",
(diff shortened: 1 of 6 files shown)


public/js/libs/firebase.js
```diff
@@ -5,6 +5,6 @@ let config = {
     projectId: "avf-dashboards",
     storageBucket: "avf-dashboards.appspot.com",
     messagingSenderId: "1063777163571",
-    appId: "1:1063777163571:web:57361756b9ecb959"
+    appId: "1:1063777163571:web:57361756b9ecb959",
```
IsaackMwenda: is this necessary

IsaackMwenda (approved):

public/js/libs/firebase.js
DanielMwendwa: No, its just from the settings in my editor

PR #257, opened 2020-07-03
Update graph_layout.js

DanielMwendwa:
I've moved the function used to add labels to graphs to `graph_layout.js`. This is also to ensure that the function will be used on other graphs I'm working i.e bar chart, stacked bar chart therefore there will be no need for repetition.


IsaackMwenda (approved):
LGTM.  Next time move the code and apply it to the relevant graphs in one pr so that we can test it.

DanielMwendwa:
> elevant graphs in one p

Noted

DanielMwendwa:
> LGTM. Next time move the code and apply it to the relevant graphs in one pr so that we can test it.

Noted

PR #262, opened 2020-07-13
remove blurry line on axis path

DanielMwendwa:
I've plotted the black axis path on top of the light grey gridlines path and reduced its stroke width to prevent the blurry effect on graph's axis

### DIFF
--- a/public/js/libs/area_chart.js
+++ b/public/js/libs/area_chart.js
@@ -50,7 +50,6 @@ export class AreaChart extends GraphLayout {
         let decimalFormatter = d3.format(".2s");
         this.yAxis = d3
             .axisLeft(this.yScale)
-            .ticks(5)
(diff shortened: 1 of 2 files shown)


public/js/libs/area_chart.js
```diff
@@ -90,6 +88,11 @@ export class AreaChart extends GraphLayout {
             .attr("class", `${this.id}Grid`)
             .call(d3.axisLeft(this.yScale).tickSize(-this.width).tickFormat(""));
 
+        // this.plot
```
IsaackMwenda: remove commented code block

IsaackMwenda (approved):

public/js/libs/area_chart.js
DanielMwendwa: updated, thanks

PR #263, opened 2020-07-14
display y limit on chart

DanielMwendwa:
Screenshot 2020-07-15 at 20 52 11


public/js/libs/area_chart.js
```diff
@@ -98,6 +99,23 @@ export class AreaChart extends GraphLayout {
             .call(d3.axisBottom(this.xScale).tickSize(-this.height).tickFormat(""));
     }
 
+    displayYLimit() {
```
IsaackMwenda, line 100: It would probably be better if we added the limit to the last tick in Y-axis. With different formatting to distinguish from the rest. Otherwise, the current design makes the graph look too busy.

IsaackMwenda (changes requested):

public/js/libs/area_chart.js
DanielMwendwa: Updated, thanks.

DanielMwendwa:
Screenshot 2020-07-23 at 12 40 02
I've made the y limit bold

IsaackMwenda (approved):

PR #267, opened 2020-08-05
update active dropdown-item background color

DanielMwendwa:
Update the background colour for active projects dropdown items when clicked to light steel blue.


IsaackMwenda (commented):
This is a trivial change. Current text change to blue is enough

DanielMwendwa:
> This is a trivial change. Current text change to blue is enough

Noted.

PR #274, opened 2020-09-10
RESOLVE: High false alert rate for last updated timestamp on dashboard graphs

### LINKED ISSUE #271: High false alert rate for last updated timestamp on dashboard graphs
e.g.
image

I've seen this quite a few times recently, but not sure exactly under which conditions it shows.

DanielMwendwa:
closes #271
I have included a function that clear all the setIntervals and invoked it where one navigates to a new page. I have also updated the time that checks for downtime for the traffic operators from 20 minutes to 30 minutes.

### DIFF
--- a/public/js/controllers/traffic_graphs_controller.js
+++ b/public/js/controllers/traffic_graphs_controller.js
@@ -1034,7 +1034,7 @@ export class TrafficGraphsController {
             // Calculate time diff bw current and lastUpdateTimeStamp
             let currentTime = new Date(),
                 difference_minutes = (currentTime.getTime() - lastUpdateTimeStamp.getTime()) / 60000;
-            if (difference_minutes > 20) {
+            if (difference_minutes > 30) {
                 d3.select("#lastUpdated").classed("text-stale-info alert alert-stale-info", true);
             } else {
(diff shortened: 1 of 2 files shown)


public/js/controllers/traffic_graphs_controller.js
```diff
@@ -1034,7 +1034,7 @@ export class TrafficGraphsController {
             // Calculate time diff bw current and lastUpdateTimeStamp
             let currentTime = new Date(),
                 difference_minutes = (currentTime.getTime() - lastUpdateTimeStamp.getTime()) / 60000;
-            if (difference_minutes > 20) {
+            if (difference_minutes > 30) {
```
IsaackMwenda: why are we increasing this? The dashboard updates every 10 mins so the sooner we know it is down the better.

IsaackMwenda (approved):

public/js/controllers/traffic_graphs_controller.js
DanielMwendwa: I changed to 30 minutes to be same as for system metrics but you are right.I have updated to 20 minutes.

PR #278, opened 2020-10-12
Update filter by active projects

DanielMwendwa:
Filtering the coding progress table by project TIS+ isn't working. I've split the keyword TIS+ on the plus sign to enable filtering with TIS only.

### DIFF
--- a/public/js/controllers/coding_progress_table_controller.js
+++ b/public/js/controllers/coding_progress_table_controller.js
@@ -220,7 +220,7 @@ export class CodingProgressTableController {
             // Filter rows
             if (tableSection == "td") {
                 if (CodingProgressTableController.keyword != "") {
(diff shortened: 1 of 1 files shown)


IsaackMwenda (commented):
Renaming the project name in active_projects to TIS seems like the most appropriate action @as2388 @DanielMwendwa

as2388:
I think the issue is I called this TIS+ in the active_projects then changed to TIS_Plus in Coda once I discovered that + still isn't safe to use everywhere. I can rename the entry in active_projects to fix and for consistency, but I think in the long term we need to use something other than the project name e.g. we could add a "CodaPrefixes": array of string to each active project. That would solve the issue that some projects have multiple prefixes, e.g. USAID_IBTCI_ and CSAP_

DanielMwendwa:
Thank you both. The best think now is to implement your suggestion of adding coda prefixes since it'll help resolve other issues as you've mentioned.

PR #279, opened 2020-10-21
Replace css framework and its dependancies with vanilla css and js

DanielMwendwa:
@IsaackMwenda @as2388, I've updated the styling of ops dashboard and removed the css framework with its dependencies i.e bootstrap js, popper js and Jquery. Reasons are first we import a lot of code which we don't use, the framework is now undergoing major changes as they are planning to release v5 thus it will be better if we've our own styling since the codebase is not that huge instead of minor upgrades occasionally and also the frame work does not support proxy browsers that is on mobile devices which might not be an issue because most of our users are not using mobile phones. I've also made the site responsive on mobile phones as was suggested earlier by a team member.


IsaackMwenda (commented):
Hi @DanielMwendwa this is a great - breaking from framework dependency. Can I however suggest we do this one framework at a time? This pr breaks so many of the UI components that I think would have been easier to identify and fix if we migrated one framework at a time. Thanks

PR #280, opened 2020-10-27
Firebase JavaScript SDK v8

DanielMwendwa:
I've updated and uncommitted three firebase js sdks(app, auth, firestore) into the project. The reason for committing libraries to the project earlier, was due to the issue of parsing long URLs to determine what the project dependencies was. For the three firebase js sdks, they are an exception, the url are concise and we'll also be updating them regularly due to the important improvements done occasionally.
Firebase JavaScript SDK v8 comes with breaking & important changes.  [Firebase JavaScript SDK release notes](https://firebase.google.com/support/release-notes/js)


public/auth.html
```diff
@@ -3,8 +3,8 @@
 <title>Login to AVF Engagement & Media Dashboard</title>
     <link rel="shortcut icon" type="image/png" href="images/favicon.png"/>
     <link rel="stylesheet" type="text/css" href="css/styles.css">
-    <script src="vendor/firebase-7.21.1/firebase-app.js"></script>
-    <script src="vendor/firebase-7.21.1/firebase-auth.js"></script>
+    <script src="https://www.gstatic.com/firebasejs/8.0.0/firebase-app.js"></script>
```
IsaackMwenda, line 6: The other reason for avoiding third-party hosted libs is that we now depend on the 3rd party to always keep the link up, I would still prefer us hosting the libraries and updating them regularly. What do you think?

IsaackMwenda (changes requested):

public/auth.html
DanielMwendwa: Since the platform is developed by google and also the fact that the old libraries are still being hosted, I'm confident the links will always be up if that is the issue.

IsaackMwenda: Thanks this also reduces our server's traffic and makes parallel loading possible.

IsaackMwenda (approved):

public/auth.html
IsaackMwenda: Thanks this also reduces our server's traffic and makes parallel loading possible.

DanielMwendwa:
Thanks Isaack

PR #282, opened 2020-11-13
Pipeline monitoring 

DanielMwendwa:


DanielMwendwa:
Hi @IsaackMwenda, I have set up pipelines metrics navigation link, page and functions that will be used to generate charts when data starts being uploaded on firestore.

DanielMwendwa:
@IsaackMwenda, I've added the table to display the pipeline name, the last start time of the pipeline, and the last successful run. I'll open a new PR for the graph.

public/js/controllers/app_controller.js
```diff
@@ -91,6 +94,21 @@ class Controller {
         });
     }
 
+    static displayPipelines() {
```
IsaackMwenda, line 97: qq - how do we restrict this to active project pipelines only?

DanielMwendwa: @IsaackMwenda, cc @as2388 I suggest we add a "project_name" field in the documents then I can update the query performed on firestore to filter out pipelines for inactive projects.

public/js/controllers/data_controller.js
```diff
@@ -74,6 +74,31 @@ export class DataController {
...
+        let offset = new Date(), 
+            offsetString, 
+            iso = d3.utcFormat("%Y-%m-%dT%H:%M:%S+%L");
+
+        offset.setDate(offset.getDate() - TIMERANGE);
+        offsetString = iso(offset);
+
+        let pipelineMetrics = [];
+        return mediadb.collection("metrics/pipelines/pipeline_logs")
+            .where("timestamp", ">", offsetString)
```
as2388, line 89: (If we're concerned strictly about the table, I don't think this is the most efficient means to retrieve data, as we're only interested in the latest documents. However, given that @DanielMwendwa you mentioned the view that shows older projects too is progressing well, no need to change here)

as2388 (approved):
Thanks @DanielMwendwa! I think what you have done so far is useful for us already if deployed. Could you consider adding issues to address in future for:
 - Filtering by active projects
 - Moving the navs so that we have a Miranda button and a Pipelines button under a Systems dropdown?

public/js/controllers/data_controller.js
DanielMwendwa: Noted, Thanks.

DanielMwendwa:
> Thanks @DanielMwendwa! I think what you have done so far is useful for us already if deployed. Could you consider adding issues to address in future for:
> 
> * Filtering by active projects
> * Moving the navs so that we have a Miranda button and a Pipelines button under a Systems dropdown?

Added

PR #283, opened 2020-12-08
Add continuous zoom ("brushing") to the traffic graphs

DanielMwendwa:
@IsaackMwenda  @as2388  plz take a look at 10 min received msg graph and kindly give me some feedback on the feature


DanielMwendwa:
@IsaackMwenda, @as2388  I've managed to add the above feature on 10 min received msg graph. Plz let me know if you'll need this feature on the 1 day msg graphs.
- I've added a clipPath on the graph: everything out of this area won't be drawn, so that the bars do not extend outside the dimensions of the graph.
- Enable brushing and zoom.
- I've updated how the bars' width are calculated so that the bars' width are updated properly when zooming the chart.
- I've updated the tooltip to work on top of the rect overlay created by the brush feature.
- I've updated the x axis to be the same as the one for system monitoring graphs.
- I've enhanced the tooltip.

public/js/controllers/traffic_graphs_controller.js
```diff
@@ -440,6 +433,165 @@ export class TrafficGraphsController {
...
+                                tooltipTranslateY = tooltipContent.length - 33;
+                            } else if (tooltipContent.length == 4) {
+                                tooltipTranslateY = tooltipContent.length - 52;
+                            } else if (tooltipContent.length == 5) {
+                                tooltipTranslateY = tooltipContent.length - 71;
+                            } else if (tooltipContent.length == 6) {
+                                tooltipTranslateY = tooltipContent.length - 90;
+                            } else if (tooltipContent.length == 7) {
+                                tooltipTranslateY = tooltipContent.length - 109;
+                            } else if (tooltipContent.length == 8) {
```
IsaackMwenda: and if it's greater than 8?

public/js/controllers/traffic_graphs_controller.js
```diff
@@ -440,6 +433,165 @@ export class TrafficGraphsController {
...
+                    let updatedDatetime = new Date(d.datetime);
+                    updatedDatetime.setMinutes(updatedDatetime.getMinutes() + 4.5);
+                    focus.attr(
+                        "transform",
+                        `translate(${x(updatedDatetime)}, ${y_total_received_sms_range(d.total_received)})`
+                    );
+
+                    let tooltipContent = [], tooltipTranslateY = 5;
+                    operators.forEach(x => {
+                        if(d.operators[x].received != 0) {
```
IsaackMwenda: please refactor to handle the  tooltipContent dynamically

IsaackMwenda (approved):
LGTM thanks @DanielMwendwa with a couple of comments
(ps - This would have been great if it was split into two PRs one for the tooltip and another for the brushing so that the diff is more clear.)

public/js/controllers/traffic_graphs_controller.js
DanielMwendwa: I had assumed the operators would be less than 8, but I get what you mean, I have updated the code block, thanks.

public/js/controllers/traffic_graphs_controller.js
DanielMwendwa: Updated, thanks.
I have refactored to handle the tooltipContent vertical alignment dynamically. The tooltip content is dynamic. I have also added comments to clarify what the code block does, thanks.

DanielMwendwa:
> LGTM thanks @DanielMwendwa with a couple of comments
> (ps - This would have been great if it was split into two PRs one for the tooltip and another for the brushing so that the diff is more clear.)

Noted, thanks.

PR #288, opened 2020-12-16
Add totals for last month/selected range to traffic graphs

DanielMwendwa:


public/js/controllers/app_controller.js
```diff
@@ -68,6 +68,13 @@ class Controller {
             );
             DataController.registerSnapshotListener(unsubscribeFunc);
         });
+        // Update and show the Metrics
+        import("./traffic_metrics_controller.js").then((module) => {
+            DataController.projectTrafficDataMetrics(
+                project,
+                module.TrafficMeticsController.updateTotals
```
IsaackMwenda:
```suggestion
                module.TrafficMetricsController.updateTotals
```

public/js/controllers/ui_controller.js
```diff
@@ -233,6 +233,9 @@ export class UIController {
                         <button class="btn btn-brown shadow-none" type="button" data-toggle="collapse" data-target="#collapseOne" aria-expanded="true" aria-controls="collapseOne">
                         Threats to Validity
                         </button>
+                        <button class="btn btn-brown shadow-none" type="button" data-toggle="collapse" data-target="#collapseTwo" aria-expanded="true" aria-controls="collapseTwo">
```
IsaackMwenda: Switch this to be the first button then Threats to Validity as the second one. Then have it collapse on default so that the metrics are visible on page load.

public/js/controllers/ui_controller.js
```diff
@@ -255,6 +258,41 @@ export class UIController {
...
+                                <div class="card responsetext">
+                                    <b>Operator metrics</b>
+                                    <table id="operator-metrics"></table>
+                                </div>
+                            </div>
+                            <div class="col">
+                                <div class="card responsetext">
+                                    <b>Project Details</b>
+                                    <div>Project Name: <span id="project-name"></span></div>
+                                    <div>AT Units Used: <span id="AT-units-used"></span></div>
```
IsaackMwenda:
```suggestion
                                    <div>Africas Talking Units Used: <span id="AT-units-used"></span></div>
```

public/js/controllers/traffic_metrics_controller.js
```diff
@@ -0,0 +1,49 @@
...
+                        TrafficMeticsController.updateTotals,
+                        selectedDates
+                    );
+                }
+            },
+        });
+
+        let totalReceived = d3.sum(data, (d) => d.total_received),
+            totalSent = d3.sum(data, (d) => d.total_sent),
+            responseRate = (Number(totalSent) / Number(totalReceived)).toFixed(2);
```
IsaackMwenda: The response rate is a bit complicated because of how Textit and flows work. e.g we sometimes send 3 - 4 messages and expect one response for them. An ideal answer for this would be response rate per flow run e.g sms ad flow.

public/js/controllers/app_controller.js
IsaackMwenda: typo

public/js/controllers/ui_controller.js
```diff
@@ -255,6 +258,41 @@ export class UIController {
                         </div>
                     </div>
                 </div>
+
+                <div id="collapseTwo" class="collapse" aria-labelledby="headingTwo" data-parent="#accordionExample">
```
IsaackMwenda: would be great if we can format this with card - shadow styles just like the rest of the dashboard sections

IsaackMwenda (approved):
Thanks @DanielMwendwa

public/js/controllers/app_controller.js
DanielMwendwa: Thanks, updated

public/js/controllers/ui_controller.js
DanielMwendwa: Thanks updated

public/js/controllers/ui_controller.js
DanielMwendwa: Thanks updated

public/js/controllers/traffic_metrics_controller.js
DanielMwendwa: I've removed the response rate from the accordion

PR #295, opened 2021-01-20
Support updating the Africa's Talking balance from sms_statistics_service.py

as2388:
Sample outputs available at avf-dashboards-test.

This lets us optionally provide an africas_talking configuration object in each active_project configuration, which contains the username of the account to look up and the url to the AT API token for that project. If the configuration is provided for a project, this will write a document containing a timestamp and balance to /metrics/africas_talking/{project_name}/{timestamp}. Note that the balance API only supports returning the balance now, so this always writes the current time rather than supporting the historical look-up that the rapid_pro updates support.

### DIFF
--- a/sms_statistics_service/src/cache.py
+++ b/sms_statistics_service/src/cache.py
@@ -51,10 +51,10 @@ class Cache(object):
                 json.dump([ap.to_dict() for ap in active_projects], f)
             return active_projects
 
-    def get_rapid_pro_token_for_project(self, project_name, google_cloud_credentials_file_path, rapid_pro_token_url):
-        cache_file_path = f"{self.cache_dir_path}/rapid_pro_tokens/{project_name}.txt"
+    def get_token_for_project(self, project_name, token_type, google_cloud_credentials_file_path, token_url):
+        cache_file_path = f"{self.cache_dir_path}/tokens/{project_name}/{token_type}.txt"
         try:
-            log.info(f"Attempting to read the Rapid Pro token for project '{project_name}' from the cache "
+            log.info(f"Attempting to read the {token_type} token for project '{project_name}' from the cache "
                      f"at '{cache_file_path}'")
             with open(cache_file_path) as f:
                 token = f.read()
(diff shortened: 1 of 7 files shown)


DanielMwendwa (approved):
Thanks @as2388

sms_statistics_service/src/firestore_wrapper.py
```diff
@@ -85,3 +88,8 @@ def update_sms_stats_batch(self, project_name, sms_stats_batch):
         assert batch_counter == 0
 
         log.info("SMS stats updated")
+
```
DanielMwendwa, line 91: Add docstrings

sms_statistics_service/src/cache.py
```diff
@@ -51,10 +51,10 @@ def get_active_projects(self, firestore_client):
...
-    def get_rapid_pro_token_for_project(self, project_name, google_cloud_credentials_file_path, rapid_pro_token_url):
-        cache_file_path = f"{self.cache_dir_path}/rapid_pro_tokens/{project_name}.txt"
+    def get_token_for_project(self, project_name, token_type, google_cloud_credentials_file_path, token_url):
+        cache_file_path = f"{self.cache_dir_path}/tokens/{project_name}/{token_type}.txt"
         try:
-            log.info(f"Attempting to read the Rapid Pro token for project '{project_name}' from the cache "
+            log.info(f"Attempting to read the {token_type} token for project '{project_name}' from the cache "
                      f"at '{cache_file_path}'")
             with open(cache_file_path) as f:
                 token = f.read()
```
DanielMwendwa, line 60: Update the log info to display the type of token in Line 61

DanielMwendwa (approved):

sms_statistics_service/src/cache.py
DanielMwendwa: Line 61

sms_statistics_service/src/firestore_wrapper.py
as2388: Added

sms_statistics_service/src/cache.py
as2388: Good catch, thanks

IsaackMwenda (approved):

PR #298, opened 2021-02-02
Display Africa's Talking balance

DanielMwendwa:


public/js/controllers/app_controller.js
```diff
@@ -75,6 +75,13 @@ class Controller {
                 module.TrafficMetricsController.updateTotals
             )
         });
+        // Update Africa's Talking balance
```
IsaackMwenda, line 75: can we import traffic_metrics_controller.js once in l72 and load all the relevant modules.

IsaackMwenda (approved):
Thanks

public/js/controllers/app_controller.js
DanielMwendwa: yes, thanks.

PR #302, opened 2021-03-11
Update The Index Page

### LINKED ISSUE #301: Put Pipelines and Miranda Navigation Links under Systems Dropdown Menu
There should be a systems dropdown menu, the pipelines & Miranda nav links as dropdown items since the pipeline metrics & Miranda metrics fall under systems metrics.

DanielMwendwa:
- I've done code clean up:- updated Html attributes especially id & classes, making sure I assign unique values to the id's and updated the CSS selectors and made other changes.
- Added systems dropdown menu
- formated code:- Break long lines into multiple lines

Since I've changed some id's and classes some links aren't working; fixed [here](https://github.com/AfricasVoices/Operations-Dashboard/pull/303)

Closes #301

### DIFF
--- a/public/css/styles.css
+++ b/public/css/styles.css
@@ -1,3 +1,4 @@
+@import url("https://fonts.googleapis.com/css2?family=Montserrat:wght@400;500;600;700&display=swap");
 .logo {
     height: 3rem;
 }
--- a/public/css/styles.css
+++ b/public/css/styles.css
@@ -15,10 +15,7 @@
     color: steelblue;
     cursor: pointer;
 }
-.traffics-link,
-.systems-link,
-.pipelines-link,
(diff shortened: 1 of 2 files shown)


as2388 (changes requested):
I think this has introduced a bug, as when I try to run this locally I get:
image
master works fine.

DanielMwendwa:
> I think this has introduced a bug, as when I try to run this locally I get:
> image
> master works fine.

This is because I am setting an event listener to an element that I have accessed with a wrong id. This after I've made the changes to class/id names. I've gone ahead and updated the code in this pr #303. I broke down this PR into two, this one just updates the navigation bar mark up.

as2388 (approved):

PR #303, opened 2021-03-11
Fix Dashboard Broken Links

DanielMwendwa:
Review #302  first
I'm not updating the hash parameters when navigating to the links under systems; this is because I'll open another PR that enables using URL parameters instead of the hash parameters e.g. 
`hostname/#traffic-GPSDD-BUNGOMA` -> `hostname/view?traffic=GPSDD-BUNGOMA`

### DIFF
--- a/public/js/controllers/app_controller.js
+++ b/public/js/controllers/app_controller.js
@@ -80,7 +80,7 @@ class Controller {
         UIController.addSystemsGraphs();
         // Update and show the Graphs
         import("./systems_graphs_controller.js").then((module) => {
-            let unsubscribeFunc = DataController.watchSystemsMetrics(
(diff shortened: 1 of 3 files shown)


public/js/controllers/app_controller.js
```diff
@@ -79,12 +76,8 @@ class Controller {
         DataController.watchATCredits(project, displayATCredits);
     }
 
-    static displaySystems() {
+    static displayServerMetrics() {
```
as2388: If you're using Miranda in the UI can you call this displayMirandaMetrics instead?

as2388 (approved):

public/js/controllers/app_controller.js
DanielMwendwa: updated, thanks.

PR #304, opened 2021-03-18
Language-Sensitive Representation of Numbers

DanielMwendwa:
Updated unique texts, Unique Texts with a label, Wrong Scheme messages, Not Coded messages columns on the coding progress table; Africas Talking Units Balance; Message metrics & Operator metrics sections in the traffic page.

### DIFF
--- a/public/js/controllers/traffic_metrics_controller.js
+++ b/public/js/controllers/traffic_metrics_controller.js
@@ -50,7 +50,8 @@ export class TrafficMetricsController {
         if (data.length) {
             let { balance } = data[0];
             let [ currency, amount ] = balance.split(" ");
-            node.text(`${Number(amount).toLocaleString("en-KE", { style: "currency", currency })}`);
+            let { language } = navigator;
+            node.text(`${Number(amount).toLocaleString(language, { style: "currency", currency })}`);
         } else {
             node.text("N/A");
         }
--- a/public/js/controllers/traffic_metrics_controller.js
+++ b/public/js/controllers/traffic_metrics_controller.js
@@ -24,8 +24,8 @@ export class TrafficMetricsController {
             totalSent = d3.sum(data, (d) => d.total_sent);
 
         d3.select("#project-name").text(projectName);
-        d3.select("#total-received").text(totalReceived);
-        d3.select("#total-sent").text(totalSent);
+        d3.select("#total-received").text(totalReceived.toLocaleString());
+        d3.select("#total-sent").text(totalSent.toLocaleString());
 
         const metricsTable = document.getElementById("operator-metrics");
         while (metricsTable.firstChild) {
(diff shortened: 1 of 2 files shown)


public/js/controllers/traffic_metrics_controller.js
```diff
@@ -34,8 +34,8 @@ export class TrafficMetricsController {
         let html = "<tr>%operator_metrics%</tr>";
         operators.forEach((operator) => {
             let tableData = `<td>${operator}</td>
-                <td>${d3.sum(data, (d) => d[`${operator}_received`])}</td>
-                <td>${d3.sum(data, (d) => d[`${operator}_sent`])}</td>`;
+                <td>${d3.sum(data, (d) => d[`${operator}_received`]).toLocaleString()}</td>
+                <td>${d3.sum(data, (d) => d[`${operator}_sent`]).toLocaleString()}</td>`;
```
as2388, line 38: Thanks, these look better

public/js/controllers/traffic_metrics_controller.js
```diff
@@ -49,7 +49,8 @@ export class TrafficMetricsController {
         let node = d3.select("#AT-units-bal");
         if (data.length) {
             let { balance } = data[0];
-            node.text(balance);
+            let [ currency, amount ] = balance.split(" ");
+            node.text(`${Number(amount).toLocaleString("en-KE", { style: "currency", currency })}`);
         } else {
             node.text("N/A");
```
as2388, line 58: Drive-by comment for another PR: can this be "N/A" only if AfricasTalking isn't present in the project configuration? If it is, and there is no data, show an error here instead. This also means if we remove africastalking from the configuration this should switch back to "N/A"

public/js/controllers/traffic_metrics_controller.js
```diff
@@ -49,7 +49,8 @@ export class TrafficMetricsController {
         let node = d3.select("#AT-units-bal");
         if (data.length) {
             let { balance } = data[0];
-            node.text(balance);
+            let [ currency, amount ] = balance.split(" ");
+            node.text(`${Number(amount).toLocaleString("en-KE", { style: "currency", currency })}`);
```
as2388: Why is this en-KE when the others are the user's default? Is there a way to pass in the default (something like navigator.language)? I think it should be user's default everywhere, for consistency.

(If not then en-KE with an explanatory comment is fine - everyone in AVF should be comfortable with this locale)

as2388 (approved):

public/js/controllers/traffic_metrics_controller.js
DanielMwendwa: Noted

public/js/controllers/traffic_metrics_controller.js
DanielMwendwa: Thought it was better because it produces the 'Ksh' sign which we are used to instead of 'KES' code but it's not a big deal. Updated thanks.

PR #312, opened 2021-04-09
Display AT balance as N/A if AT isn't Present in Project Configuration

DanielMwendwa:
@as2388 below are changes as requested earlier from this [comment](https://github.com/AfricasVoices/Operations-Dashboard/pull/304#discussion_r598784911)


public/js/controllers/data_controller.js
```diff
@@ -146,7 +147,21 @@ export class DataController {
...
     }
 
+    static watchProjectsWithATconfig(activeProjects) {
+        let projectsWithATconfig = new Set();
+        activeProjects.forEach((project) => {
+            if (project.hasOwnProperty("africas_talking")) {
+                projectsWithATconfig.add(project.project_name);
+            }
+        });
+        DataController.projectsWithATconfig = [...projectsWithATconfig];
```
as2388: I think this approach is slightly more complicated than it need be, and risks being error prone if you ever end up forgetting to keep this `projectsWithATconfig` property up to date.
  Instead of loading the active_projects, synchronising them with a list of names that match a condition, passing the project_name around, then checking that list, why not just pass the full project object around? Then in the watchATCredits function you can just say `if (project.hasOwnProperty("africas_talking"))...`.

as2388 (changes requested):

public/js/controllers/data_controller.js
DanielMwendwa: Updated, thanks. I've not been able to get the project object but I've resolved the issue. I'll work on using browser storage to store some of the info like active projects objects, this will help to reduce the number of reads we're making.

public/js/controllers/data_controller.js
```diff
@@ -147,14 +147,28 @@ export class DataController {
...
+
+                let ATCredits = [];
+                return mediadb
+                    .collection(`metrics/africas_talking/${projectName}`)
+                    .orderBy("datetime", "desc")
+                    .limit(1)
+                    .onSnapshot((res) => {
+                        // Update data every time it changes in firestore
+                        DataController.updateData(res, ATCredits);
+                        onChange(ATCredits);
```
as2388: Given that you're requesting up to 1 result, I think it would be slightly cleaner if you passed that result here rather than the list the result lives in

as2388 (approved):
Thanks, approved with one small, optional suggestion

public/js/controllers/data_controller.js
DanielMwendwa: Thanks for the suggestion, updated.

PR #313, opened 2021-04-15
Adds Duration Column on Pipeline Monitoring Table

DanielMwendwa:


public/js/controllers/pipelines_controller.js
```diff
@@ -27,6 +27,12 @@ export class PipelinesController {
                     break;
                 }
             }
+
+            pipelineProgress["Duration"] = "-";
+            if (pipelineProgress["Last Successful Run"] > pipelineProgress["Last Start Time"]) {
```
IsaackMwenda: Check if the run id for both event is the same too. We could be running the pipeline in two different machines e.g locally and in Miranda. The local pipeline is running successfully but Miranda is failing. Cross-checking run id  helps us detect failure for each pipeline run.

IsaackMwenda (approved):
Good progress thanks.

public/js/controllers/pipelines_controller.js
DanielMwendwa: Thanks, updated.

PR #327, opened 2021-10-13
Toggle between stacked bar charts and regular bar charts

DanielMwendwa:


public/js/libs/helper.js
```diff
@@ -0,0 +1,129 @@
+const plotSingle = (graph, layer, transition = true) => {
```
as2388:
```suggestion
const plotSingleOperator = (graph, layer, transition = true) => {
```

public/js/libs/helper.js
```diff
@@ -0,0 +1,129 @@
...
+    graph.y_orig = []; // to store original y-posn
+    layer.selectAll("rect")._groups[graph.idx].forEach(function (d, i, n) {
+        // Get height and y posn of base bar and selected bar
+        let h_keep = d3.select(d).attr("height");
+        let y_keep = d3.select(d).attr("y");
+        // Store y_base in array to restore plot
+        graph.y_orig.push(y_keep);
+
+        let h_base, y_base;
+        layer.selectAll("rect")._groups[0].forEach(function (d, i, n) {
```
as2388: Is there a way of doing this without poking around at the graph's private internal state (_groups)?

public/js/libs/helper.js
as2388: Can you make this filename more descriptive? What do these helpers help with?

public/js/libs/helper.js
```diff
@@ -0,0 +1,129 @@
...
+        if (!transition) {
+            d3.select(d).attr("y", y_new);
+        } else {
+            d3.select(d).transition().ease(d3.easeBounce).duration(500).delay(150).attr("y", y_new);
+        }
+    });
+    return graph;
+};
+
+const restorePlot = (graph, layer) => {
```
as2388: Is there a name you can give that describes the type of restore?

With a name like restorePlot I'd expect either an arg containing the original data or an implementation that redraws from original data. Maybe something like showAllOperators?

public/js/libs/helper.js
```diff
@@ -0,0 +1,129 @@
...
+    let legendHovered = d3.select(target).datum().replace(/\s/g, "");
+    if (graph.activeLink === "0") d3.select(target).style("cursor", "pointer");
+    else {
+        if (graph.activeLink === legendHovered) {
+            d3.select(target).style("cursor", "pointer");
+        } else d3.select(target).style("cursor", "auto");
+    }
+};
+
+const cellClickHandler = (target, graph, layer) => {
```
as2388: Can you rename these handlers in terms of what they actually do rather than where they should be called?

as2388 (approved):
Functionality is great, just a few comments on names.

If you have a time, it would be great to be able to change selected graph without having to deselect the selected one first (e.g. to be able to go straight from golis to hormuud), in another PR.

public/js/libs/helper.js
DanielMwendwa: This was a bit challenging but I've managed to update it, thanks.

public/js/libs/helper.js
DanielMwendwa: I've renamed the file to `ctrl_stacked_bars.js`, couldn't find a better name that is shorter and precise.

public/js/libs/helper.js
DanielMwendwa: I've renamed the function to `showAllOperators`, thanks.

public/js/libs/helper.js
DanielMwendwa: I've renamed `cellOverHandler` to `indicateSelecteableLegends` and `cellClickHandler` to `ctrlGraphByLegendSelections`, thanks.

DanielMwendwa:
> Functionality is great, just a few comments on names.
> 
> If you have a time, it would be great to be able to change selected graph without having to deselect the selected one first (e.g. to be able to go straight from golis to hormuud), in another PR.

Ok, will work on that later, thanks.

public/js/controllers/traffic_graphs_controller.js
```diff
@@ -219,7 +251,7 @@ export class TrafficGraphsController {
         let receivedSliderStep = d3
             .sliderLeft()
             .min(0)
-            .max(yLimitReceived + (0.2 * yLimitReceived))
+            .max(yLimitReceived + 0.2 * yLimitReceived)
```
IsaackMwenda:
```suggestion
            .max(yLimitReceived + (0.2 * yLimitReceived))
```
use nested parenthesis here to separate the math equation

public/js/controllers/traffic_graphs_controller.js
```diff
@@ -228,16 +260,13 @@ export class TrafficGraphsController {
...
-            .append("g")
-            .attr("class", "sentSlider")
-            .attr("transform", `translate(${-45},${0})`);
+        total_sent_sms_graph.append("g").attr("class", "sentSlider").attr("transform", `translate(${-45},${0})`);
 
         // Step
         let sentSliderStep = d3
             .sliderLeft()
             .min(0)
-            .max(yLimitSent + (0.2 * yLimitSent))
```
IsaackMwenda, line 240: ditto

public/js/controllers/traffic_graphs_controller.js
```diff
@@ -283,7 +312,14 @@ export class TrafficGraphsController {
             .shapeWidth(12)
             .orient("vertical")
             .scale(colorReceived)
-            .labels(operators);
+            .labels(operators)
+            .on("cellover", function () {
```
IsaackMwenda, line 316:
```suggestion
            .on("cellhover", function () {
```

IsaackMwenda: For quick monitoring, it would be great to draw the selected network on cellHover even without clicking and reset to all otherwise.

public/js/controllers/traffic_graphs_controller.js
```diff
@@ -298,7 +334,14 @@ export class TrafficGraphsController {
             .shapeWidth(12)
             .orient("vertical")
             .scale(colorSent)
-            .labels(operators);
+            .labels(operators)
+            .on("cellover", function () {
```
IsaackMwenda, line 338:
```suggestion
            .on("cellover", function () {
```
I think we should refer to this as legends and not cells so on `.on("legendhover", function () {`

IsaackMwenda (approved):
Looks  good with a couple comments and suggestions thanks @DanielMwendwa .

public/js/controllers/traffic_graphs_controller.js
DanielMwendwa: Updated thanks.

public/js/controllers/traffic_graphs_controller.js
DanielMwendwa: We're using [d3-legend](https://d3-legend.susielu.com/). There are three custom event types we can bind to the legend: "cellover", "cellout", and "cellclick".

public/js/controllers/traffic_graphs_controller.js
DanielMwendwa: This will not be efficient since the graphs require time to transition or else the graphs become distorted. Just try to click one legend quickly simultaneously, the graph will become distorted. later I'll implement debouncing and throttling to solve the issue completely.

